PR #167, opened 2026-07-08
Move team management into hierarchy

Rothermal:
## Summary
- add /team/manage as the team management route and point Team Info manage links there
- keep /teams focused on cloud team creation/listing while preserving roster, invite, and merge workflows on Team Manage
- update regression coverage for the new Team Manage flow

## Verification
- pnpm.cmd test -- src/lib/teamInfo.test.ts
- pnpm.cmd lint (existing Fast Refresh warnings in context files)
- pnpm.cmd build (existing chunk-size warning)

### DIFF
--- a/docs/REGRESSION_TESTING.md
+++ b/docs/REGRESSION_TESTING.md
@@ -68,6 +68,8 @@ High-level test scripts for features built so far. Use these to sanity-check aft
 | 4.8 | Reload → Teams | Same teams and roster (from Supabase) |
 | 4.9 | Teams → tap a team card's primary name area | Opens `/team?teamId=<id>`; Team Info shows display name, season, sport, record, roster count, game count, and links back to Season Stats, Team Stats, and Manage Team |
 | 4.10 | Teams → tap **Manage** on a team card | Opens `/team/manage?teamId=<id>` with roster and member management for that team |
+| 4.10b | Open legacy `/teams?teamId=<id>` link | Redirects to `/team/manage?teamId=<id>` |
+| 4.10c | Open `/team/manage` without a team id or with an invalid id | Shows Team unavailable state; does not auto-select or manage another team |
 | 4.11 | Team Info → switch **Overview / Roster / Schedule** segments | Segment content changes in place; URL stays `/team?teamId=<id>`; Manage links open `/team/manage?teamId=<id>` |
 | 4.12 | Team Info → Overview | Shows stat links, roster preview, schedule preview, recent results, tournaments, and read-only team members when data exists |
 | 4.13 | Team Info → Roster | Shows the full active roster and read-only member list; player rows link to `/player-info?playerId=<id>&teamId=<id>` |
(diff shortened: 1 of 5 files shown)


Rothermal:
## Code review (TI-7)

Happy-path split looks good: `/teams` = list/create, `/team/manage` = roster/members/merge; `teamManagementPath` callers (Team Info Manage, list Manage, create → manage) and regression updates line up. Reusing `Teams.tsx` with a mode flag is a reasonable staging step vs embedding management into Team Roster/Info yet (plan/design Phase 8 still describe that fuller move — worth calling out as intentional follow-up).

### Important before merge

**1. `/team/manage` without a valid `teamId` silently manages the first team**

In `src/pages/Teams.tsx` (~156–161), selection still falls back to `loadedTeams[0]` when `requestedTeamId` is missing or not in the loaded list. That was tolerable when manage lived on the list page; on a dedicated manage route it means `#/team/manage` (or a bad id) can edit the wrong team with no empty/error state. Sibling Team Info routes show “Missing team” instead.

**Suggested fix:** On the management route, if `!requestedTeamId` or the id is not in `loadedTeams`, show an empty/error state or `navigate('/teams', { replace: true })` — do not auto-select the first team. Back with no selected team should go to `/teams`, not `/admin`.

**2. Old deep links `/teams?teamId=` no longer open management**

List UI now ignores `teamId` for roster/members (hidden behind `isManagementRoute`), so bookmarks / muscle memory / any leftover links to `/teams?teamId=X` only get a leftover selection highlight at best. Prefer a redirect:

`/teams?teamId=X` → `/team/manage?teamId=X`

(or explicitly clear list selection styling and document the break).

---

Not blockers: invite banner also on manage (plan allows preserve-on-`/teams`); scorer roster mutation UI gating is pre-existing; full embed into Team Roster/Info can be a later slice if this route extraction is the accepted TI-7 staging.

Rothermal:
Validated the review comments and both were real:

- /team/manage without a valid teamId could fall back to the first team, so I changed management mode to clear stale selections, avoid auto-selecting a team, and show a Team unavailable state with Back to Cloud Teams.
- legacy /teams?teamId=<id> links no longer opened management after the route split, so they now redirect to /team/manage?teamId=<id>.

Also updated the regression checklist for both cases.

Verification:
- pnpm.cmd test -- src/lib/teamInfo.test.ts
- pnpm.cmd build (existing chunk-size warning only)

Rothermal:
## Re-review (after `ebe03b5`)

Prior Important items look fixed:

1. **Invalid/missing manage `teamId`** — manage route no longer falls back to `loadedTeams[0]`; shows **Team unavailable** and keeps roster/members gated on a real `selectedTeam`. Back goes to `/teams` when there is no selected team.
2. **Legacy `/teams?teamId=`** — replace-redirects to `/team/manage?teamId=`. Regression 4.10b / 4.10c cover both cases.

No new Important issues from the fix commit. Happy path (create → manage, list Manage, Team Info Manage, invites, merge) still looks intact.

**Ready to merge** from a review standpoint (TI-7 as management-route extraction / staging; fuller embed into Team Roster/Info remains a later slice per plan/design).

PR #168, opened 2026-07-08
Add team start game navigation

Rothermal:
## Summary
- add Team Info → Start Game and route it to /setup?teamId=<id>
- teach Game Setup to preselect a requested cloud team, including direct-link sport lookup and invalid-team empty state
- route team-context back actions from Leaderboard, Team Stats, and Tournament Stats toward Team Info while preserving secondary stats links
- update regression coverage for the Start Game and back-link flows

## Verification
- pnpm.cmd test -- src/lib/teamInfo.test.ts
- pnpm.cmd lint (existing Fast Refresh warnings in context files)
- pnpm.cmd build (existing chunk-size warning)

### DIFF
--- a/docs/REGRESSION_TESTING.md
+++ b/docs/REGRESSION_TESTING.md
@@ -81,9 +81,12 @@ High-level test scripts for features built so far. Use these to sanity-check aft
 | 4.18 | Team Schedule → tap a scheduled/live/final game | Opens `/game-info?gameId=<id>&teamId=<id>`; shows game details, status/score, and stat leaders when resolved stats exist |
 | 4.19 | Game Info → View full summary on a finalized game | Hydrates the cloud game through the existing flow and opens `/summary`; pending local sync still blocks hydration |
 | 4.20 | Team Roster → tap a player | Opens `/player-info?playerId=<id>&teamId=<id>` with **Player Info**, season totals, game log, career link, and **Back to Team** returning to `/team?teamId=<id>` |
-| 4.21 | Leaderboard → tap a player row | Existing `/player?teamId=<id>&playerId=<id>&seasonId=<id>` route still opens **Player Profile** and backs to Leaderboard |
+| 4.21 | Team Info → Season Stats → back arrow | Opens Leaderboard with that team selected; back arrow returns to `/team?teamId=<id>` |
 | 4.22 | Team Info hero → season name | Opens `/team/season?seasonId=<id>` with season name, sport, team count, and teams in that season |
 | 4.23 | Season Info → tap a team name / Season Stats | Team name opens `/team?teamId=<id>`; Season Stats opens Leaderboard with `seasonId` and `teamId` set |
+| 4.24 | Team Info → Start Game | Opens `/setup?teamId=<id>` with the team's sport and existing team preselected |
+| 4.25 | Open `/setup?teamId=<id>` directly | Loads the team sport when possible and preselects the team; invalid/inaccessible team shows Team unavailable and does not auto-select another team |
+| 4.26 | Team Stats / Tournament Stats / Game Info / Player Info → back action | Returns to `/team?teamId=<id>` when opened with team context |
 
 ---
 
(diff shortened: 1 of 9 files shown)


Rothermal:
## Code review (TI-8)

Team Info → Start Game looks solid: discard confirm matches SportSelect, `SET_SPORT` resets via `createInitialState`, then cloud sync + `/setup?teamId=`. GameSetup preselect (requested team wins, invalid team does not fall through to first team) and Team/Tournament Stats secondary links are good. Games.tsx skip is fine for this slice.

### Critical — fix before merge

**Leaderboard ← from home now goes to Team Info**

Home opens `/leaderboard` with no params (`SportSelect`). On load, Leaderboard always writes `teamId` into the URL (`pushLeaderboardParams` when teams load). This PR’s back logic treats any `teamId` as Team Info context:

```ts
const hasTeamContext = Boolean(teamIdFromUrl)
// ← → teamInfoPath(selectedTeamId) when hasTeamContext
```

So: Home → Season Stats → ← incorrectly lands on `/team?teamId=…` instead of `/`. That breaks the plan’s “keep global shortcuts available from home.”

**Suggested fix:** Don’t treat auto-injected URL params as Team Info intent. e.g. only back to Team Info when entry was from Team Info (`from=team` / navigation state), or stop auto-writing `teamId` when the user arrived with an empty query, or back to `/` unless a sticky “came from team” flag is set. Retest Home → Season Stats → ←.

### Important

1. **Direct `/setup?teamId=` with a different sport already in state** — sport-load effect bails when `sport` is truthy (`GameSetup.tsx`). A basketball local sport + soccer team link can mark the team “unavailable” even though it exists. If `requestedTeamId` is present, resolve that team’s sport (and `SET_SPORT` when it differs).

2. **Direct `/setup?teamId=` can wipe an active game with no confirm** — Team Info confirms before discard; the sport-load `SET_SPORT` path does not. Mirror the SportSelect/Team Info confirm when an active game would be reset.

3. **Regression 4.21 rewrite** — old “Leaderboard player row → Player Profile → back to Leaderboard” coverage was replaced. §9.3 still covers the tap; restore a back-path row (e.g. 4.21b) so that isn’t lost from the Team Info suite.

---

Happy-path Start Game + Team/Tournament Stats backs look mergeable once the Leaderboard home regression (and preferably the direct-setup sport/discard issues) are fixed.

Rothermal:
Validated the review comments and updated the branch.

Valid fixes made:
- Leaderboard now uses an explicit rom=team query marker for Team Info-origin links. Auto-injected 	eamId params from Home → Season Stats no longer make the back arrow go to Team Info.
- /setup?teamId= now resolves the requested team's sport even when a different sport is already in game state.
- If resolving that requested team sport would reset an active game, setup now mirrors the existing discard confirmation and returns home if canceled.
- Restored the regression row for Leaderboard player row → Player Profile → back to Leaderboard.

Verification:
- pnpm.cmd test -- src/lib/teamInfo.test.ts
- pnpm.cmd build (existing chunk-size warning only)

Rothermal:
## Re-review (after `a08da5d`)

Prior issues look fixed:

1. **Critical — Leaderboard home back** — `from=team` is now explicit (`teamLeaderboardPath(..., true)` from Team Info / Team Stats; Leaderboard only backs to Team Info when `from=team`). Home → Season Stats → ← returns to `/`. Param is preserved across season/team dropdown updates.
2. **Important — direct `/setup?teamId=` sport mismatch** — sport is always resolved from the requested team; `SET_SPORT` runs when it differs.
3. **Important — discard confirm on that path** — confirm before wiping an active game; cancel navigates home.
4. **Regression 4.21b** — Player Profile back coverage restored.

### Remaining (non-blocking)

- **Season Info / Team Manage → Season Stats** still open Leaderboard without `from=team` (`SeasonInfo.tsx`, `Teams.tsx` raw `/leaderboard?teamId=`), so ← goes home rather than Team Info. Optional polish: pass `fromTeam` (or use `teamLeaderboardPath`) for hierarchy consistency.
- **Direct `/setup?teamId=` when sport already matches** does not confirm/reset an active game (only the mismatch path does). Team Info → Start Game still always resets. Edge case only.

**Ready to merge** from a review standpoint.

PR #182, opened 2026-07-12
Add multi-game parking P0

Rothermal:
## Summary
- add local multi-game parking storage with manifest, per-game records, legacy migration, and multi-sport summaries
- expose park/resume/discard/start-new actions through GameContext and wire home/setup/team/summary flows to park instead of discard
- update docs/agent notes to mark P0 shipped and P1 sync queue as next

## Verification
- pnpm.cmd test -- src/lib/gameParking.test.ts
- pnpm.cmd build
- pnpm.cmd lint (passes with existing react-refresh warnings in context files)


Rothermal:
## Code review — multi-game parking P0

**Verdict: not merge-ready** until the migration failure path is fixed. Core P0 shape is good; a few gaps break “park instead of overwrite.”

### What works
- Storage model matches the plan: manifest + `statkeeper_game:{id}` + legacy mirror
- Home / setup / team / summary use park APIs instead of wipe
- Finalize clears only the active parked record
- Sync still targets the mounted game; stale fingerprint guard still drops late merges onto the wrong mounted state
- Unit tests cover happy-path migration, multi-sport rows, park, activate, empty discard

---

### Critical

**1. Legacy migration can delete the only copy on write failure** (`src/lib/gameParking.ts` ~173–204)

If `setItem` / `writeManifest` throws (e.g. quota), the `catch` calls `removeLegacyMirror()`. That can wipe `statkeeper_game` after a partial write.

**Fix:** Only remove legacy after both record and manifest writes succeed; on failure, leave legacy intact (and roll back any new record key if needed). Add a test that simulates `QuotaExceededError`.

---

### Important

**2. Cloud open paths still overwrite the active local slot**

`Games.tsx`, `GameInfo.tsx`, `PlayerProfile.tsx`, `CareerStats.tsx` still `HYDRATE_STATE` into the current active id. The persist effect then `saveActiveGameState`s over that same local record — so a fully synced active game A is replaced locally when opening cloud game B. Plan acceptance is “park instead of overwrite.”

**Fix:** Park (or `beginNew` + hydrate into a new local id) before cloud open, with the same confirm pattern as SportSelect.

**3. Mid-sync park can drop cloud ids for the parked game**

`GameContext` treats in-flight sync as stale after switch and does not merge sync results into the parked record. Resuming that game can call `ensureGame` again → duplicate cloud rows. More likely now that park-during-sync is first-class UX.

**Mitigate:** Flush/await before park/start, or merge sync results into the snapshot’s `localGameId` record.

**4. Global `statkeeper_pending_sync` is not per local game**

Acceptable as known P0 debt if documented; otherwise dirty A can lose retry behavior while working on B.

**5. Acceptance tests incomplete**

Missing: park A → start B → resume A (stats/shots), migration failure must not wipe legacy, cloud-open must not clobber active local.

---

### Minor
- Double `setParkedGames` in `parkCurrentGame`
- `beginNewActiveParkedGame` can register an active id with no record until first save
- `listParkedGames` includes active; UI filters it out — naming is a bit confusing
- Incomplete setup rows can clutter the parked list

---

### Suggested merge bar
1. Fix Critical #1 (+ test)
2. Either fix Important #2 in this PR, or explicitly defer cloud-open park as a named follow-up and update the plan acceptance notes
3. Call out #3/#4 as accepted P0 risk with P1 ownership

Rothermal:
Reviewed the reported issues and pushed follow-up commit 6df6885.

Validated/fixed:
- Legacy migration write failure was valid. migrateLegacyGameStorage now preserves statkeeper_game if manifest/record migration fails and rolls back only the new per-game record. Added a quota-style regression test.
- Cloud-open overwrite path was valid. Games, GameInfo, PlayerProfile, and CareerStats now open cloud snapshots through openGameSnapshot, which saves/parks the current active local game and hydrates the cloud game into a fresh local active id.

Accepted P0 debt / documented for P1:
- In-flight sync results are still active-game scoped and can be dropped as stale after switching. This needs captured-localGameId merge or flush semantics in P1.
- statkeeper_pending_sync is still global and should become per-parked-record dirty state in P1.

Verification:
- pnpm.cmd test -- src/lib/gameParking.test.ts (151 tests passed)
- pnpm.cmd build
- pnpm.cmd lint (passes with existing fast-refresh warnings)

Rothermal:
## Re-review after `6df6885`

**Verdict: merge-ready for P0**, with previously flagged sync items accepted as documented P0 debt.

### Resolved

**Critical #1 — legacy migration wipe** ✅  
`migrateLegacyGameStorage` no longer calls `removeLegacyMirror()` on write failure; it rolls back only the new per-game record and leaves `statkeeper_game` intact. Quota-style regression test covers this.

**Important #2 — cloud open overwrite** ✅  
`Games`, `GameInfo`, `PlayerProfile`, and `CareerStats` now go through `openGameSnapshot`, which saves the current active game, starts a new local active id, then hydrates the cloud snapshot. Confirm prompts match the SportSelect park pattern.

**Important #3 / #4 — mid-sync stale merge + global pending sync** ✅ (deferred)  
Called out under **Known P0 debt** in `docs/PLAN_MULTI_GAME_PARKING.md` for P1. Acceptable for this PR.

### Remaining (non-blocking)

- `parkCurrentGame` still does a redundant double `setParkedGames`
- `beginNewActiveParkedGame` can briefly register an active id with no record until the next save
- Quota failure test’s follow-up `listParkedGames` succeeds on retry (throw-once storage) — fine, but means first paint can briefly show no active game until something re-reads parking
- Opening the same cloud game twice creates a second local parked copy (minor clutter)

### Verification
Locally: `pnpm test -- src/lib/gameParking.test.ts` (6 passed), `pnpm build` passed.

Rothermal:
Reviewed the new re-review comment and pushed small cleanup commit 788d194.

Validated and updated:
- The redundant double setParkedGames in parkCurrentGame was valid. I removed the extra state write while preserving the explicit save-before-park behavior.

Validated but not changing in P0:
- eginNewActiveParkedGame can briefly set an active id before a record exists. I agree this is technically true, but it is transient during normal React flow and self-heals on reload by discarding the empty active id. No user data is at risk, so I do not think this needs to block P0.
- The quota test uses throw-once storage, so a later listParkedGames can retry migration and succeed. That matches the intended recovery behavior; first failed read preserves legacy, next read can migrate once storage works.
- Opening the same cloud game twice can create duplicate local parked copies. Valid minor clutter, but not data loss. I would handle this as a future polish/dedupe pass rather than expanding P0 further.

Verification after cleanup:
- pnpm.cmd test -- src/lib/gameParking.test.ts (151 tests passed)
- pnpm.cmd build
- pnpm.cmd lint (passes with existing fast-refresh warnings)

Rothermal:
## Re-review after `788d194`

**Verdict: still merge-ready for P0.**

### Update
- Redundant double `setParkedGames` in `parkCurrentGame` is fixed — save-before-park is preserved; list refresh comes from `parkActiveGame` only.

### Remaining minors (agree: do not block P0)
- Brief active id before first record — transient; reload discards empty active
- Quota throw-once test then successful retry — intended recovery
- Duplicate local copies of the same cloud game — polish/dedupe later

No new Critical/Important findings. Sync-queue debt remains correctly scoped to P1.

Rothermal:
## Fresh code review (HEAD `788d194`, post-merge)

**Context:** PR is already merged to `stattracker`. This is a full pass on the final branch, including prior fixes. Prior Critical (migration wipe) and Important (cloud-open overwrite via `openGameSnapshot`) remain resolved.

**Verdict: request follow-up hotfix before relying on P0 in production.** New findings below were not covered in earlier reviews.

---

### Still good
- Manifest + per-game records + fail-safe legacy migration (+ quota test)
- Manual cloud opens park-then-open via `openGameSnapshot` (Games / GameInfo / PlayerProfile / CareerStats)
- Home/setup/team/summary park flows; sport-labeled parked list
- Finalize uses `clearActiveParkedGame`
- Documented P0 debt for mid-sync stale merge + global pending sync

---

### Critical

**1. Auth / ownership races can wipe or bleed parked games**  
`GameContext.tsx` ~728–756 + `gameParking.ts` `readManifest` / `saveActiveGameState`

Persist effect (`[state, userId]`) runs before the user-switch reset and writes with the **new** `userId` while React state may still be the previous user’s game.

- **A → B without going through `signOut`’s `clearPersistedGameStorage`:** `readManifest(B)` sees owner A ≠ B → `clearAllParkedGames()`, then may briefly re-save A’s in-memory game under B before reset discards it → **A’s parked set wiped**.
- **Unexpected `user` → `null` (session expiry, not `signOut`):** persist can rewrite `ownerId` to `null`. Mismatch clear only runs when **both** ids are truthy (`parsed.ownerId && ownerId && …`), so a later user B can **inherit** null-owned parked games. Intentional `signOut` clears storage first (good); other null transitions do not.

**Fix direction:** Don’t persist across `userId` changes until after reset; never overwrite a real `ownerId` with `null`; on account switch, clear/namespace by owner without a stale-state save; hide Resume/Discard when logged out if parking is account-scoped.

---

### Important

**2. Auto cloud resume still overwrites the active local slot in place**  
`GameContext.tsx` ~866–868 vs `openGameSnapshot`

Manual opens park-first. Auto-hydrate still `HYDRATE_STATE`s into the current active id. `shouldDeferCloudResumeHydration` only protects dirty / unsynced / no-`gameId` locals — a **fully synced** active game A can be replaced by preferred/latest cloud game B, and the persist effect writes B into **A’s `localGameId`** (A disappears from parking).

**Fix:** If an active local game exists (or local `cloudGameId` ≠ hydrate target), use `openGameSnapshot` / park-first, or skip auto-hydrate when a persistable active local already exists.

**3. `removeLegacyMirror` clears global `statkeeper_pending_sync` on park / new game**  
`gameParking.ts` ~154–157, 296–307

Parking/starting new calls `removeLegacyMirror()`, which removes the durable pending-sync flag. Documented P0 debt is “global pending”; **clearing it on park** can drop the only retry signal for a dirty offline parked game.

**4. `beginNewActiveParkedGame` can leave orphan `gameIds` with no record**  
Crash/quota between `beginNew` and first `saveActiveGameState` → ghost ids (filtered from UI, never cleaned). Prefer adding to `gameIds` only after first successful record write.

---

### Minor
- Inconsistent `hasActiveGame` checks (players-only vs `gameInfo || players`) across SportSelect vs cloud-open pages
- Final cloud games opened via `openGameSnapshot` cannot persist (`gameStatus === 'final'` discards) — memory-only review; document or allow a review slot
- `startNewGame` / `parkCurrentGame` don’t sync `stateRef` before dispatch (narrow race vs `openGameSnapshot` / resume)

---

### Suggested follow-up
1. Hotfix Critical #1 (auth/ownership) — highest priority  
2. Important #2 (auto-hydrate park-first) in the same hotfix if practical  
3. Important #3/#4 can ship with that hotfix or early P1

Happy to review a follow-up PR against these items.

PR #184, opened 2026-07-12
feat: harden multi-game cloud sync p2

Rothermal:
## Summary
- resolve player ids and team roster links before inserting a new cloud games row during snapshot sync
- best-effort delete a just-created in-progress cloud game if downstream stat, shot chart, or team-placeholder writes fail before the client persists the cloud id
- add regression coverage for sync ordering and rollback behavior
- update multi-game parking docs, README, AGENTS, overview, integration checklist, and regression script to mark P2 shipped and clarify remaining follow-ups

## Verification
- pnpm.cmd test
- pnpm.cmd build


Rothermal:
## Code review — multi-game cloud sync hardening P2

**Verdict: merge-ready for P2 scope**, with a few Important/Minor notes. This PR correctly narrows the orphan-`games` window; it does not claim (and should not claim) full transactional sync or historical orphan cleanup.

### What works
- Player/roster resolution (`ensurePlayerId` loop) now runs **before** `ensureGame` insert — failures there no longer create an in-progress cloud game with no stats
- `EnsuredGame.created` gates rollback so existing `cloudSync.gameId` updates are not deleted on child write failure
- Best-effort `deleteNewlyCreatedGameAfterFailedSync` scoped to `id` + `created_by` + `status=in_progress`
- Child writes (stats → shot chart → team placeholder link) wrapped so insert+child failure attempts rollback before rethrow
- Unit test asserts roster ops precede `games.insert`, and `games.delete` runs after `game_stats` failure
- Docs/README/AGENTS/regression §3a accurately describe shipped P2 vs remaining follow-ups (no full RPC transaction; no historical cleanup)

### Important (non-blocking if documented)

**1. Crash window still exists (accepted P2 limit)**  
If the process dies after `games.insert` and before the `catch` rollback, an orphan in-progress row can remain. Plan already says full transactional/RPC sync is future work — keep that explicit so “P2 shipped” is not read as “orphans impossible.”

**2. Shot-chart “soft” outcomes do not trigger rollback**  
`syncShotChartToCloud` can return `synced_partial` / `skipped_*` without throwing. Stats may already be written; the new game row is kept. That is probably correct for incomplete-hydration cases, but it means “child write failure → delete” only covers **thrown** errors, not partial chart sync. Worth a one-line comment near the try block or in the plan.

**3. Rollback delete does not inspect Supabase `{ error }`**  
Failures are swallowed (`.catch(() => {})` and no error check on delete). Fine for best-effort, but a failed delete (RLS/FK) silently leaves the orphan the PR tries to avoid. Consider logging via existing `logClientSyncError` when delete fails (follow-up OK).

**4. Test coverage is thin for the rollback matrix**  
Only the `game_stats` failure path is covered. Suggest (follow-up) cases for: shot-chart throw, `linkGameTeamPlaceholderIds` throw, and `created: false` (existing gameId) **must not** delete.

### Minor
- Empty `statRows` short-circuits stats upsert (pre-existing); a brand-new game with no stats yet can still be inserted then rolled back only if a later child throws — edge case for pre-stat sync
- P0 follow-ups (**auth/ownership races**, **auto-hydrate overwrite**) are untouched here — still open; do not treat this PR as resolving them

### Plan alignment

| P2 claim | Status |
|---|---|
| Resolve roster/players before new `games` insert | Met |
| Best-effort delete just-created in-progress game on child write failure | Met |
| Historical orphan cleanup tooling | Explicitly not shipped |
| Full transactional / idempotent cloud write | Explicitly not shipped |

### Suggested merge bar
Ship as-is for P2. Optional same-PR polish: log failed rollbacks; add the two extra hardening tests. File/keep open the P0 auth + auto-hydrate hotfixes separately.

Rothermal:
Reviewed the PR feedback and pushed 827f046. I treated the merge-ready review as valid optional polish and updated the PR: rollback delete now inspects Supabase errors and logs failed rollback attempts through client_sync_errors with rollback/original-error context; the sync code now explicitly comments that only thrown child-write failures trigger rollback, while shot-chart partial/skipped modes intentionally keep the game row; the plan now calls out the accepted crash window and soft shot-chart outcomes; cloudSyncHardening coverage now includes stats failure, shot-chart throw, team-placeholder link throw, existing-game no-delete, and rollback-delete logging. Remaining notes are accepted P2 limits/follow-ups: full transactional/RPC sync, historical orphan cleanup, and the older P0 auth/hydrate items. Verification: pnpm.cmd test (161 passed), pnpm.cmd build, pnpm.cmd lint (existing Fast Refresh warnings only).

Rothermal:
## Re-review (`827f046`)

**Verdict: still merge-ready.** Follow-up addressed the Important polish from the first pass. Local check: `cloudSyncHardening.test.ts` (5/5) and `pnpm build` pass.

### Addressed
- **Rollback observability:** `deleteNewlyCreatedGameAfterFailedSync` now returns the PostgREST error message; failures go through `logClientSyncError` with `bypassThrottle` + `rollbackGameId` / `originalSyncError` context (and a test asserts `client_sync_errors.insert`).
- **Soft vs hard shot-chart outcomes:** Inline comment + plan Phase 2 text make clear only *thrown* child failures roll back; partial/skipped chart modes keep the row on purpose.
- **Crash window:** Documented as accepted P2 limit (not a full DB transaction).
- **Coverage:** Stats fail, shot-chart throw, placeholder-link throw, existing-game no-delete, rollback-delete logging.

### Remaining (accepted / out of scope for this PR)
- Browser crash between `games.insert` and rollback can still orphan an `in_progress` row — P2 follow-up / transactional RPC territory.
- Open P0 items from earlier parking work (auth/`userId` persist races; auto cloud hydrate overwriting the active local slot) remain separate from this hardening PR.

Ship it.

PR #185, opened 2026-07-12
feat: add multi-game storage guardrails p3

Rothermal:
## Summary
- add local parked-game storage guardrails: 12-game cap, quota/max-count errors, and storage estimate
- add Settings export/import for local parked games
- thread parking storage failures through GameContext and guard navigation paths that would otherwise proceed after a failed save
- add gameParking coverage for export/import, invalid import, max count, and quota errors
- update multi-game docs/README/AGENTS/regression notes as P3a shipped, with IndexedDB left as follow-up

## Verification
- pnpm.cmd test
- pnpm.cmd build


Rothermal:
## Code review — P3a storage guardrails

**Verdict: needs changes before merge.** Cap / quota / export-import core look solid and match the plan P3a slice. Navigation + error surfacing is incomplete relative to this PR’s claim to guard failed saves and show user-visible storage errors. `gameParking.test.ts` 15/15 pass on the branch.

---

### Critical

1. **`GameSummary` still navigates after a failed park** — `src/pages/GameSummary.tsx` `handleNewGame` (~525–531)

```ts
parkCurrentGame()
navigate('/')
```

`parkCurrentGame()` now returns `boolean`, but Summary ignores it and always leaves. `newGameError` UI is never set. This file is not in the PR diff at all — a clear miss vs “guard navigation paths that would otherwise proceed after a failed save.”

**Fix:** `if (!parkCurrentGame()) { setNewGameError(...parkingError or storage message); return }`

2. **Live autosave quota failures are effectively invisible on the tracker** — `GameContext` sets `parkingError` on autosave failure, but only `SportSelect` renders it. Regression **3a.11** expects a visible storage/quota message; on Game Tracker the user can keep logging into memory that never persists.

**Fix:** Surface `parkingError` in a shared layout / tracker chrome (not only home).

---

### Important

3. **Cloud open paths abort navigate but show no error** — `Games.tsx`, `GameInfo.tsx`, `CareerStats.tsx`, `PlayerProfile.tsx`: correctly stop on `openGameSnapshot === false`, but never copy `parkingError` into the page `error` state. Failure looks like a no-op after the spinner clears.

4. **Specific storage messages dropped on some start-game paths** — `TeamInfo.tsx` / `GameSetup.tsx` hardcode `"Parked games could not be saved on this device."` instead of using `parkingError` / `parkedGameStorageErrorMessage` (cap vs quota differ).

5. **`openGameSnapshot` can partially mutate storage before failing** — `GameContext.openGameSnapshot`: save current → `beginNewActiveParkedGame` (new active id in manifest, legacy mirror cleared) → save snapshot. If the last `safeSetItem` throws, React state is unchanged but the manifest already points at an empty new active id. Next autosave can write the old in-memory game into that slot (duplicate / odd cap accounting). Prefer not advancing `activeLocalGameId` until the new snapshot write succeeds, or roll the manifest back on catch.

6. **Import overwrite + weak validation (accepted follow-up, still worth noting)** — same `localGameId` silently replaces blobs; Admin skip copy always says “at the parked-game limit” even for schema-invalid rows; envelope check only (`version` + `records`), not a shallow `GameState` shape. Plan already defers conflict UI — OK for P3a if documented, but easy footgun.

---

### Minor

7. Settings KB display floors at 1 KB (`Math.max(1, …)`).
8. `isQuotaError` only checks modern DOMException names; some WebKit paths historically used `code === 22`.
9. Test gaps: import overwrite, import-at-cap skip reasons, mid-import quota, and caller guards (especially `GameSummary`).

---

### What looks good

- Cap enforced on create paths (`beginNewActiveParkedGame`, first-time `saveActiveGameState`, etc.).
- Quota wrapped via `safeSetItem` with clear `ParkedGameStorageError` messages.
- Most start/open callers check boolean returns; SportSelect sport-select + parked list error banner are correct.
- Export/import rebinds `ownerId` to the current user; Settings estimate + docs/regression 3a.8–3a.11 updated; IndexedDB correctly deferred.

---

### Out of scope (not blockers for this PR)

Prior P0 auth/`userId` persist races and auto cloud-hydrate overwrite remain separate. IndexedDB + import conflict UI are already listed as P3 follow-up.

**Merge bar:** Fix Critical #1 and #2 (ideally Important #3–#5). After that, P3a is close to merge-ready.

Rothermal:
Updated in a1033a7.

Validated and fixed:
- GameSummary now honors parkCurrentGame() returning false and stays on Summary when parking fails.
- Game Tracker now surfaces the context parkingError, so live autosave quota/cap failures are visible while tracking.
- Cloud open paths now render parkingError on Games, Game Info, Player Profile, and Career Stats instead of failing as a silent no-op after openGameSnapshot() returns false.
- Team Info and Game Setup no longer replace storage-specific failures with the generic parked-game message; they render the context parkingError so cap vs quota remains specific.
- openGameSnapshot now rolls back a newly-created active local game id and restores the prior active parked game if the snapshot save fails after beginNewActiveParkedGame().
- isQuotaError now accepts older code === 22 quota exceptions in addition to modern DOMException names.

Deferred intentionally:
- Import conflict UI / overwrite policy and deeper imported GameState shape validation are still P3 follow-up scope, not required for this P3a guardrail slice.
- KB display flooring is cosmetic and not blocking.

Verification:
- pnpm.cmd test -- src/lib/gameParking.test.ts (Vitest ran all 165 tests): pass
- pnpm.cmd build: pass
- pnpm.cmd lint: pass with existing Fast Refresh warnings only

Rothermal:
## Re-review (`a1033a7`)

**Verdict: merge-ready.** Follow-up addresses the Critical and Important merge-bar items from the prior review. Local check: `gameParking.test.ts` 15/15 and `pnpm build` pass.

### Addressed

- **Critical #1 — GameSummary:** `handleNewGame` now checks `parkCurrentGame()` and surfaces `parkingError` instead of always navigating.
- **Critical #2 — Tracker visibility:** `GameTracker` shows an amber `parkingError` banner under the scoreboard (covers live autosave / 3a.11).
- **Important #3 — Cloud open paths:** `Games` / `GameInfo` / `CareerStats` / `PlayerProfile` render `parkingError` (or fold it into existing error UI).
- **Important #4 — Cap vs quota copy:** `TeamInfo` / `GameSetup` use `parkingError` instead of a hardcoded generic string.
- **Important #5 — Partial `openGameSnapshot`:** On failure after `beginNewActiveParkedGame`, discards the new slot and reactivates the previous active id; refreshes parked list from storage.
- **Minor — WebKit quota:** `isQuotaError` also treats `code === 22`.

### Remaining (non-blocking)

- Import overwrite / shallow `gameState` validation / skip-reason copy — already deferred as P3 follow-up.
- `parkingError` is still page-local (e.g. Players / other mid-flow screens won’t show an autosave failure). Tracker + home + start/open paths cover the high-risk surfaces; a layout banner can wait.
- No automated tests for the new caller guards / openGameSnapshot rollback path.

Ship it.

PR #186, opened 2026-07-12
feat: polish parked game imports p3b

Rothermal:
## Summary
- carry forward the P3 follow-up planning doc and mark P3b decisions as settled
- make Settings parked-game import a parked-only merge that keeps existing local ids, skips invalid/at-cap rows with reason-specific counts, and never switches the active game
- roll back records written by an import batch when a quota/manifest write failure occurs
- update Settings import/storage messaging plus regression and agent docs

## Verification
- pnpm.cmd test -- src/lib/gameParking.test.ts (Vitest ran 169 tests)
- pnpm.cmd build
- pnpm.cmd lint (passes with existing Fast Refresh warnings only)


Rothermal:
## Code review — P3b import polish

**Verdict: merge-ready.** Import merge semantics match the settled P3b decisions in the plan: parked-only, keep-existing on id conflict, shallow validation with reason-specific skips, partial import at cap, and batch rollback when the final manifest write (or a mid-batch `safeSetItem`) fails. Local check: `gameParking.test.ts` 19/19 and `pnpm build` pass.

### What looks good

- **No overwrite / no active switch:** skip if `gameIds.includes` or `readRecord`; `writeManifest` always keeps `manifest.activeLocalGameId`; no legacy mirror write on import.
- **Skip taxonomy:** invalid → existing → at-cap, with separate counters wired through Admin (`existing kept` / `over limit` / `invalid row`).
- **Rollback:** `writtenLocalGameIds` cleaned on catch; prior manifest left untouched when `writeManifest` throws.
- **Docs:** P3b marked shipped; P4 IndexedDB / P5 cloud ops cleanly deferred; regression 3a.9–3a.14 updated; storage display no longer floors everything to `1 KB`.

### Minor (non-blocking)

1. **Coverage gaps:** no test that a pre-existing active game (and legacy mirror) stay put when the payload nominates a different `activeLocalGameId`, or that pre-existing parked rows survive a failed-import rollback. Code looks correct; worth a couple asserts later.
2. **In-file duplicate ids** count as `skippedExisting` (“existing game(s) kept”) even when the “existing” row was written earlier in the same batch — safe, slightly imprecise copy.
3. **Admin always reloads**, including all-skipped (`imported === 0`) results.
4. **Whitespace `localGameId`:** trim is required for validity, but the stored id is untrimmed.

None of these are merge blockers. Ship it.

PR #214, opened 2026-07-18
SOC-2C: add soccer live match controls

Rothermal:
## Summary\n- replace the soccer staging surface with the SOC-2C live match tracker\n- add anchored clock, period, lineup, substitution, role, direction, rules, participant, and end/reopen controls\n- add revisioned event correction history with delete/restore and projection diagnostics\n- complete SOC-2 documentation and add the SOC-2C regression matrix\n\n## Scope\n- remains development-only and local-only until later soccer phases\n- does not add soccer scoring, field events, cloud event sync, or the final summary\n\n## Verification\n- pnpm test (58 files, 355 tests)\n- pnpm build\n- pnpm lint (0 errors; 3 existing Fast Refresh warnings)\n- git diff --cached --check\n- local Vite smoke check: HTTP 200


Rothermal:
## Code review — PR #214 (SOC-2C live match controls)

**Verdict: needs changes** before merge.

CI green. The live tracker slice is in good shape overall: Result-checked helpers before `HYDRATE_STATE`, incomplete-history UI lock + fail-closed appends, DEV gate intact, no aggregate sync leak, and REGRESSION §11d covers the right manual surface. Three Important issues remain around substitution window defaults, running-clock correction vs minutes, and double-submit races.

### What works

- Staging replaced by `SoccerGameTracker`; production still redirects soccer `/game`.
- Live helpers (`live.ts`) use `addGameEvent(s)` / update / delete / restore and surface `.ok`; UI hydrates only on success.
- Incomplete projection: diagnostics banner, live controls disabled, History tab forced; appends still roll back if bypassed.
- Clock display is UI-timer only (no per-second reducer writes); atomic period end / next-period batches look correct.
- End/reopen, GK/sub-limit rejections via projector, basketball path unchanged.
- Docs mark SOC-2 complete; AGENTS/overview/README aligned.

### Important

1. **Halftime defaults for every period break.**  
   `SubstitutionForm` sets `halftime` from `projection.status === 'period_break'`, so H2→ET / ET1→ET2 / any break defaults to “no window consumed.” Plan decision 39: only **halftime** changes skip a window. Default should be true only after the first regulation segment (true HT), otherwise false (user may still override).

2. **Running-clock correction can inflate participation minutes.**  
   `adjustSoccerClock` while running does pause → adjust → start. Pause closes open intervals at the *pre-correction* elapsed time; the stopped adjust path then rewinds the clock without rewriting those closed intervals. Backward corrections leave `totalActiveMs` ahead of the corrected timeline. Prefer adjusting on the running path (or explicitly rewind/clamp open/closed intervals) so minutes follow the corrected clock.

3. **No in-flight guard on live applies.**  
   Handlers close over `state` and `HYDRATE_STATE` the result. A second tap before re-render can rebuild from the old stream and overwrite the first hydrate. Disable primary actions while applying, or serialize applies through a single-flight queue.

### Minor

- Clock correction dialog seeds MM:SS once (`useState`) and always edits match time even when the header shows period time — easy to mis-enter.
- Bench “Role” still offered for `left` players while the form drops them.
- Sub action is menu-only (plan mentions focused sub + role on lineup rows) — product polish.
- `SoccerMatchHistory` uses `React.Dispatch` without an explicit type import (tsc ok today).

### Merge bar

Fix Important #1 (halftime default) and #2 (running adjust vs participation). Prefer #3 before merge as well — live taps will hit it. After that, re-review.

Rothermal:
Validated all reported items against the projector and live orchestration.\n\nImplemented in e6f7f05:\n\n- Important 1: halftime now defaults only at the break after the first regulation segment. Later regulation/extra-time breaks default to a normal substitution window; the recorder can still override.\n- Important 2: a running clock correction is now one in-place clock-adjusted event. The projector advances to the event's from-time, reanchors at the corrected time, and keeps active participation intervals open. Added coverage proving a 60s -> 45s correction yields 55s of both match and participant time 10s later.\n- Important 3: live hydration now has a synchronous ref-backed single-flight guard plus busy UI state. A second stale submission is rejected until the new GameState render clears the guard.\n- Minor: period-clock mode now explicitly says corrections use cumulative match time and displays the current period value as context.\n- Minor: departed participants no longer expose a Role action that cannot target them.\n- Minor: React Dispatch/SetStateAction types are now explicit imports.\n\nDeferred:\n- Row-level substitution shortcuts remain UI polish for SOC-3's field/lineup work. The existing substitution action is complete and this does not affect correctness or SOC-2 event coverage.\n\nVerification:\n- pnpm test: 58 files / 355 tests passed\n- pnpm build: passed\n- pnpm lint: 0 errors; 3 pre-existing Fast Refresh warnings\n- git diff --check: clean

Rothermal:
## Code review — PR #214 (re-review)

**Verdict: merge-ready**

CI green. Prior Important items are addressed in `e6f7f05`.

### Prior findings — resolved

1. **Halftime default:** `isSoccerHalftimeBreak` requires period break after exactly the first regulation segment; covered for HT vs later breaks.
2. **Running-clock correction:** single `clock_adjusted` on the running path; projector advances the anchor from `fromElapsedMs` before apply so open intervals are not closed at the pre-correction watermark; test asserts participation follows the corrected clock.
3. **Double-submit:** `applyingRef` + full-screen busy overlay; dialogs/history take `busy` and `onApply` returns whether hydrate was accepted; cleared when `state` updates.

Also cleaned up: per-period clock correction hint, Role hidden for `left` players, explicit `Dispatch` import in history.

### Residual minors (non-blocking)

None that should delay merge.

### Merge bar

Safe to merge as the SOC-2C live controls / correction history slice. Soccer remains DEV-only and local-only until later phases.

PR #219, opened 2026-07-18
SOC-3C: add soccer Timeline and field review

Rothermal:
## Summary
- add side-colored, outcome-shaped field markers with side and period filters plus marker-driven correction
- unify live capture, historical attacking-event entry, and revision editing in one shared sheet
- replace History with a filtered Timeline and add scoring history with signed, reason-required adjustments
- complete SOC-3 documentation and hand the roadmap to SOC-4

## Verification
- pnpm test -- --run (60 files, 383 tests)
- pnpm build
- responsive browser QA at 390x844 and 1280x900, including marker correction, historical goal entry, score adjustment, Timeline filters, and horizontal-overflow checks


Rothermal:
## Code review — PR #219 (SOC-3C)

**Verdict: needs changes** — Timeline, markers, historical add, scoring history, and shared editor mostly deliver SOC-3C, but shot ↔ own-goal toggling in edit mode can append a second live event instead of revising.

CI green. Docs/README hand off to SOC-4; regression §11f looks complete.

### What works

- Field markers: side/period filters, outcome shapes, side colors; drawn inside the flipped SVG so they rotate with the pitch; marker tap opens the shared sheet
- Timeline replaces History: All / Attacking / Match Control, newest-first, removed collapsed by default, period-local `MM:SS`
- `recordHistoricalSoccer*` / `reviseSoccer*` / score adjust APIs with `soccerPeriodTimings` bounds; historical add does not move the live clock
- Shared capture sheet supports live / historical / edit, any-participant historical picks, role-at-time GK default, location re-pick
- Score buttons open Scoring Timeline; signed ±1 + required reason; goals/own goals/adjustments listed
- Incomplete projection still locks live capture and Add missed; attacking edit/restore path covered in live tests
- Basketball untouched

---

### Critical

**1. Edit-mode Own goal toggle can append instead of revise** (`SoccerShotCaptureDialog.tsx` `save`)

`GameEventEditableFields` cannot change `eventType`. Save branches only revise when the stored type already matches the checkbox:

- Edit a `soccer.shot`, check **Own goal** → falls through to `recordSoccerOwnGoal` (live append) while the shot remains
- Edit a `soccer.own_goal`, uncheck **Own goal** → falls through to `recordSoccerShot` while the own goal remains

During an active period this silently doubles scoring/history. After period end the live path often errors, but the control still implies a type change.

Lock the Own goal control in `mode === 'edit'` (copy: remove + re-add if they need the other type), or implement an explicit delete+append conversion. Do not fall through from edit/historical into the live record helpers when types disagree.

---

### Important

**2. Edit/historical save fallthrough to live helpers is unsafe generally**

Even with the checkbox locked, `mode === 'edit' | 'historical'` should fail closed if `moment` is missing or the event type does not match—never call `recordSoccerShot` / `recordSoccerOwnGoal`. Today `timingInvalid` disables the button, but the fallthrough remains a footgun.

---

### Minor

1. Score dialog `onApply` closes the sheet only when `!inspection.complete` (repair path); success correctly returns to the list via `setEditing(null)`—slightly surprising but OK.
2. Revising a `+1` adjustment into `-1` is not preflight-blocked the way append `-1` at score 0 is; projector still fail-closes / diagnostics.
3. Marker filter state is session-local (fine; not in the park contract).

---

### Merge bar

Fix Critical **#1** (and harden Important **#2** fallthrough) before merge. Rest of SOC-3C looks ready.

Rothermal:
Reviewed and updated in 6d148ed.

The reported shot/own-goal toggle was not currently reachable through the UI because edit mode already hides the Own goal control. The underlying fallthrough was still valid as a defensive concern, so capture save routing now fails closed through an explicit operation resolver:
- live mode can only call live record helpers
- historical mode requires a recorded match moment and can only append historically
- edit mode requires a matching existing event type and can only revise
- type mismatch instructs the recorder to remove and add a replacement; it cannot append a second event

Added focused regression coverage for matching revisions, mismatched types, historical timing, and live routing. Full suite passes: 60 files / 384 tests. Lint remains 0 errors with the 3 existing Fast Refresh warnings; production build passes.

The minor observations remain intentional: invalid score-adjustment revisions are preserved for Timeline diagnostics, marker filters are session-local, and successful score-adjustment edits return to the scoring list.

Rothermal:
## Code review — PR #219 (re-review)

**Verdict: merge-ready**

CI green. `6d148ed` closes the Critical edit fallthrough and the Important edit/historical→live footgun.

### Prior findings — resolved

1. **Shot ↔ own-goal edit:** `resolveSoccerCaptureSaveOperation` rejects type changes in edit mode with an explicit remove-and-replace message; no live append of a second scoring event.
2. **Fallthrough hardened:** historical/edit require a moment; save dispatches only `revise` / `record_historical` / `record_live` from the resolver. Unit coverage for live/historical/revise success and type/timing failure cases.

### What works

- Unchanged from the prior review: markers, Timeline filters, historical add, scoring timeline, shared editor, docs/§11f, basketball isolation.

### Residual minors (non-blocking)

- Own goal checkbox remains toggleable in edit mode; save fail-closes with copy. Optional polish: disable the control when `mode === 'edit'` so the constraint is visible before save.

### Merge bar

Safe to merge. SOC-3C is complete for handoff to SOC-4.

PR #227, opened 2026-07-23
SOC-5C: add independent recorder resolution

Rothermal:
## Summary
- allow authorized soccer recorders to bind independent event streams to the same cloud game while preserving shared participant identity
- persist one stable provisional primary recorder, expose compact recorder health, and add audited owner/admin primary selection
- add isolated read-only recorder projections and a primary-stream cloud review without blending streams into active game state
- update Cloud Games resume/review behavior plus SOC-5C architecture, access, and regression documentation

## Verification
- pnpm test (70 files, 461 tests)
- pnpm build
- git diff --cached --check

## Deployment
- apply Supabase migration 045_soccer_recorder_resolution.sql after migrations 043 and 044

## Deferred
- primary locking, finalization readiness, and canonical publication remain SOC-5D
- production soccer release and final summary surfaces remain SOC-6


Rothermal:
## Code review: SOC-5C independent recorder resolution

**Verdict: Needs changes** before merge.

Stream isolation, owner/admin primary selection, personal-game rejection, and `SECURITY DEFINER`/`search_path` look intentional and mostly solid. The blockers are Cloud Games bootstrap treating load failure / empty cloud as “start a new stream,” and the live default-primary trigger locking first-wins instead of creator preference.

### What works

- Read-only other-recorder projections stay in local React state; tests assert the active stream is unchanged.
- `set_soccer_primary_recorder` requires owner/admin (or personal owner), current conflict-free checkpoint; audit + access-audit event.
- `bind_soccer_event_game_v3` existing-game path does not rewrite `games.created_by` / `client_local_game_id`; personal multi-recorder rejected.
- Event loads/upserts remain recorder-scoped (`recorded_by`); same-recorder merge path unchanged.
- Primary tables are SELECT-only for clients; new functions set `search_path = public`.
- Finalized soccer → `/soccer/review`; DEV-gated review route; CI green.

### Critical

1. **Cloud Games conflates load failure / empty cloud with “start independent recorder”**

   ```359:380:src/pages/Games.tsx
         let soccerGame = await loadSoccerCloudGameById(userId, game.id).catch(err => {
           setError(...)
           return null
         })
         if (!soccerGame) {
           ...
           const startIndependent = window.confirm(
             'Start your own independent recorder stream for this game? ...'
           )
           ...
           soccerGame = await createSoccerIndependentRecorderState(...)
   ```

   `loadSoccerCloudGameById` returns `null` for non-creators with **zero cloud events**, and any throw (bad projection, transport) is also forced to `null`.

   **Failures**
   - Recorder B already recording locally (kickoff + goals, not yet synced) → Cloud Games open → cloud empty for B → confirm → **fresh kickoff** parks/orphans prior local work against the same `gameId`.
   - Existing cloud stream that fails to project → error swallowed → offered a **second** kickoff; next sync collides with prior opening events.
   - No check of parked/active locals with `cloudSync.gameId === game.id` before offering independent start.

   **Fix:** Distinguish (a) load/projection error → show error, do not offer kickoff; (b) parked/active local binding for this cloud game → resume it; (c) true first-time second recorder with empty own stream → confirm independent start. Never map `.catch → null → kickoff`.

### Important

2. **Default primary is first current checkpoint, not creator-preferring**

   Plan §Provisional primary / `effective_soccer_primary_recorder` prefer a healthy **creator** checkpoint, then earliest healthy. Backfill does that (`order by (recorded_by = created_by) desc`).

   Live trigger `assign_default_soccer_primary_recorder` inserts whoever first gets a current checkpoint (`on conflict do nothing`). Once a row exists, `effective_*` returns it forever.

   **Failure:** B syncs before A → B becomes durable default primary; A’s later healthy sync does not take it back. Breaks `11l.1` if B wins the race. Prefer: only auto-insert when recorder is creator, or do not persist defaults and rely on `effective_*` until an explicit selection.

3. **Shared `game_participants` last-writer-wins on every recorder sync**

   Identity remap is protected, but `display_name` / `jersey_number` / `snapshot` (includes current status/role) are overwritten from the syncing recorder’s projection (`045` ~478–493). Divergent subs/roles → last sync wins shared metadata. Prefer creator-only metadata updates, or omit live status/role from the shared snapshot.

4. **Creator re-sync no longer refreshes game header fields**

   v3 existing-game path never updates `games` metadata (opponent/date/name). Safe for binding identity; accidental freeze vs v2. Document or allow creator-only header updates.

5. **CI does not cover the dangerous multi-recorder paths** — migration tests are string-match; `recorders.test.ts` covers parsing/isolated projection/kickoff shape, not Games null→kickoff or primary race.

### Minor

6. Tracker attention badge if *any* recorder is non-current (noisy).
7. Error set in `.catch` can stick after a successful independent start.
8. Concurrent owner/admin primary changes: last write wins, both audited (acceptable).

### Merge bar

Block on **Critical #1**. Strongly fix **Important #2** before calling SOC-5C done vs the plan/`11l.1`. Apply `045` only after those land.

Rothermal:
Review follow-up pushed in cde5bde.

1. **Critical - fixed.** Cloud Games now resolves a matching active/parked local binding before any cloud request, prefers active then dirty local work, and resumes that slot. Cloud load/projection errors propagate and stop the flow; only a successful null/empty recorder result can offer a new independent kickoff. Missing game/setup now throws instead of returning empty.
2. **Primary race - fixed.** Automatic primary rows/triggers were removed. Until an owner/admin explicitly selects a primary, the resolver dynamically prefers a healthy creator checkpoint, then the earliest healthy checkpoint. Explicit selections remain durable if later stale.
3. **Shared participant last-writer-wins - fixed.** Shared snapshots no longer contain current role/status and remain first-write setup/identity metadata. Non-creators cannot overwrite shared names/numbers; recorder live state remains event-derived.
4. **Creator header refresh - fixed.** Existing-game v3 binds now allow creator-only updates to tracked team, opponent, competition, and date. Other recorders cannot rewrite the header.
5. **Coverage - fixed.** Added direct Cloud Games source-decision tests for active/dirty local precedence, cloud vs empty distinction, and load-error propagation, plus updated migration/static contracts and manual matrix cases. Full suite: 71 files / 466 tests.
6. **Any-recorder attention indicator - retained intentionally.** The compact row represents cross-recorder health, so an unhealthy non-primary stream remains useful manager-visible context. SOC-5D can refine finalization-specific severity without hiding the stream condition now.
7. **Sticky error - fixed by #1.** A failed load can no longer continue into successful independent creation; successful true-empty creation begins from a cleared error state.
8. **Concurrent primary writes - accepted.** Last write wins and every change remains audited, as noted.

Verification: production build passed; lint has 0 errors and the same 3 existing Fast Refresh warnings; git diff check is clean.

Rothermal:
## Re-review: SOC-5C (`cde5bde`)

**Verdict: Merge-ready** for SOC-5C.

`fix: harden soccer recorder opening` addresses the prior blockers. CI green.

### Prior findings

| # | Severity | Status |
|---|---|---|
| 1 | Critical — load failure / empty cloud → independent kickoff | **Fixed** — `resolveSoccerRecorderOpenSource` separates local / cloud / empty; load errors propagate; Games resumes parked/active bindings first; covered by `cloudOpen.test.ts` + `11l.13–14` |
| 2 | Important — first-checkpoint default primary persisted | **Fixed** — default-assign trigger + backfill removed; `effective_soccer_primary_recorder` prefers healthy creator, then earliest; only explicit selections persist |
| 3 | Important — shared participant LWW (names/snapshot) | **Fixed** — creator-only name/number updates; snapshot frozen on conflict; client no longer writes live status/role into shared snapshot |
| 4 | Important — creator header freeze on v3 re-bind | **Fixed** — creator can refresh game header fields |
| 5 | Important — no CI on dangerous open path | **Addressed** — unit tests for local precedence, empty vs cloud, error propagation; migration contracts updated |

### Residual (non-blocking)

1. Tracker attention badge if any peer stream is non-current remains noisy.
2. Concurrent owner/admin primary changes remain last-write-wins with dual audit (acceptable for provisional SOC-5C).
3. Migration contract tests are still largely SQL string-match — fine for this slice given the new open-path unit coverage.

### Merge bar

Ship when ready. Apply migration `045_soccer_recorder_resolution.sql` after 043/044. SOC-5D still owns locking / finalization readiness / canonical publication.

PR #234, opened 2026-07-24
SOC-6B3: add soccer field summary review

Rothermal:
Implements the SOC-6B3 Field tab for soccer Summary. Adds pure field-review derivation with normalized and original orientation, combined side/family/participant/period filters, deterministic marker clustering, filter-aware unknown-location review, marker detail, and authority-aware correction through shared located-event editors. Updates navigation, regression coverage, and roadmap status. Validation: 513 tests pass; lint has 0 errors and only the 3 existing Fast Refresh warnings; production build passes.


Rothermal:
## Code review — SOC-6B3 Field review

Reviewed the full diff plus the new `summaryField.ts`, `SoccerFieldReview.tsx`, and `SoccerLocatedEventEditor.tsx` against plan §6 (Field View) and the §9 SOC-6B3 verification list. Clean, well-tested slice that stays inside the 6B3 boundary.

### Strengths
- **`SoccerLocatedEventEditor` is a clean shared extraction.** `SoccerTimeline.editEvent` now delegates located events to it (removing the duplicated draft-construction + `incidentKind` logic), and Field review reuses the same component — matching plan §6 "extract shared editor orchestration; do not fork validation." Behavior is preserved, incl. shootout cards correctly falling through to the generic editor (`isSoccerLocatedEditableEvent` excludes `card` in the shootout period, with a test).
- **Edit authority reuses the SOC-6B2 hard-gated path.** `onApply={applyTimelineResult}` + `canEditSoccerSummaryField` (local && editable) means cloud primary / canonical / other-recording stay read-only with no new mutation path — consistent with the SOC-6 authority matrix.
- **Deterministic clustering.** `clusterSoccerMarkerPoints` now position-sorts input and id-sorts each cluster, with a test proving order-independence. Cluster/marker/unknown selections all feed the detail sheet in oldest-first `compareGameEvents` order.
- **Strong pure-layer tests** covering classification (incl. blocked shots in both Attack + Defense), combined filters, unlocated-count-after-filters, both orientation directions, lifecycle/shootout exclusion, period-scope options, and edit authority.
- Classification maps exactly to plan §6 families; unlocated events stay in `events` (authoritative for the count/list) while only `locatedEvents` become markers — plan's "omit only their field marker."

### Plan conformance — one item worth confirming
**Unknown-location count opens the detail sheet, not a Timeline handoff.** Plan §6 says "opens Timeline with equivalent filters *when possible* and *otherwise* opens the matching event list in a detail sheet," and §9 lists "Timeline handoff" as a verification item. The impl always opens the detail sheet (11q.6 codifies this). I believe this is actually spec-correct: SOC-6B2 Timeline supports only *family* filters — no side/participant/period — so an "equivalent filter" handoff generally isn't possible, making the detail sheet the prescribed fallback. Worth a one-line confirmation that this is the intended reading rather than a dropped requirement, since §9 names it explicitly.

### Minor nits (non-blocking)
1. **Extra-time period-aggregate redundancy.** `fieldPeriodOptions` gates the *Regulation* aggregate on `regulation.length > 1` but the *Extra Time* aggregate on `extraTime.length > 0`. A single started extra-time segment therefore yields both an "Extra Time" aggregate and an identical "Extra Time 1" individual option (same event set). The test encodes this. Consider mirroring the `> 1` guard for consistency, or leave it if intentional.
2. **`Segmented` double-labels for screen readers** — it renders both an `sr-only` `<p>{label}</p>` and `aria-label={label}` on the `role="group"`. Drop one.
3. The `participant` reset, `period` reset, and `selectedIds` prune effects each call `setState` with a fresh array whenever `review` recomputes. Traced these — no infinite loops (none feed back into the `review` memo deps), just occasional benign extra renders. No action needed; noting it was checked.

### Verdict
No blocking issues. Shootout stays hidden for 6B4, lifecycle/score-adjustments correctly produce no markers, and unlocated events stay authoritative. The only ask is author confirmation on the intentional detail-sheet-over-Timeline-handoff reading of §6.

Rothermal:
Reviewed and validated each point. The unknown-location detail sheet is intentional: Field can combine side, participant, multiple families, and period, while Timeline cannot represent that complete filter state, so the plan's matching-detail fallback applies. I clarified that acceptance wording in b675842. I also removed the duplicate screen-reader label and avoided replacing selectedIds when pruning makes no change. I am retaining the Extra Time aggregate because the plan explicitly requires semantic Regulation/Extra Time aggregates alongside individual periods; in one-period formats those scopes can currently contain the same events but remain distinct stable choices. Focused tests pass, lint has 0 errors with the same 3 existing warnings, and the production build passes.

Rothermal:
## Follow-up review — `b675842b2 refine soccer field review behavior`

The update cleanly addresses everything from my earlier review.

**Main item (detail-sheet vs. Timeline handoff) — resolved by clarifying the plan.** §9 now reads "filter-aware unknown count *with equivalent Timeline handoff when representable, otherwise the matching detail-list fallback*." That confirms the intended reading: Timeline's family-only filters can't represent an equivalent side/participant/period filter, so the detail sheet is the correct fallback. 👍

**Nit #2 (Segmented double-label)** — the redundant `sr-only <p>{label}</p>` is removed; the group keeps `aria-label={label}`, so screen readers get a single label. ✔

**Nit #3 (selectedIds prune extra render)** — now returns the same `current` reference when nothing was pruned:
```tsx
const next = current.filter(id => review.events.some(item => item.event.id === id))
return next.length === current.length ? current : next
```
Since `filter` only removes, length-equality ⟺ no removal, so the state update (and re-render) is correctly skipped unless something actually changed. ✔

**Nit #1 (extra-time aggregate redundancy)** — consciously left as-is, which is fine (I'd flagged it as optional).

No new issues; changes are minimal and safe. LGTM — good to merge from my side.

PR #235, opened 2026-07-24
SOC-6B4: add soccer shootout summary review

Rothermal:
Completes SOC-6B detailed match review with a conditional Shootout tab. Adds pure round pairing for normal series, early decisions, retakes, forfeits, and sudden death; stable anonymous slots; game-scoped kicker and goalkeeper summaries; read-only attempt detail; and suspended/abandoned state handling. Shootout values remain isolated from normal score, Players rates, and clean sheets. Updates URL handling, regression coverage, and roadmap status so SOC-6C is next. Validation: 520 tests pass; lint has 0 errors and only the 3 existing Fast Refresh warnings; production build passes.


Rothermal:
## Code review — SOC-6B4 Shootout review

Reviewed the full diff plus the new `summaryShootout.ts`, `SoccerShootoutReview.tsx`, and tests against plan §7 (Shootout View) and the §9 SOC-6B4 verification list. This is the final 6B slice, and it lands cleanly.

### Strengths
- **Strong isolation, test-verified.** `summaryShootout` reads only `projection.shootout` and never touches `sideTotals`/`participantStats`; a test clones those and asserts they're unchanged while `review.score` returns the shootout score. Matches plan §7 "never enter normal score, Players categories/rates, or clean sheets."
- **Read-only for every authority, incl. local.** The component takes only `source` — no `onApply`/edit path — and the attempt detail directs corrections to the owned tracker. Exactly plan §7's "must not add a second direct cloud mutation path." (11r.9)
- **Anonymous slots stay distinct.** Kicker/keeper summaries key on `teamSide:kickerKey` (e.g. `anonymous:1` vs `anonymous:2`), never on display label — test confirms "Opponent slot 1"/"Opponent slot 2" don't merge. (Plan §7, 11r.6)
- **Retakes/sudden-death correct.** `advances===false` retakes are excluded from `attempts`/`attemptsFaced` but preserved beneath the advancing attempt as "Retake - did not advance"; sudden-death labels count from the `initialKicksPerSide` boundary. All tested. (11r.3, 11r.5)
- **Conditional visibility** gated on `healthy && projection.shootout`, with a deep-link redirect to Overview when no shootout exists — matches SOC-6A's URL contract and plan §7 (not shown merely because rules allowed one).

### Minor items (non-blocking)
1. **Kicker-table side label shows raw `teamSide`.** `KickerRow` renders `{kicker.teamSide}` ("tracked"/"opponent") while the goalkeeper summary and score header use the actual `trackedName`/`opponentName`. Inconsistent — the kicker table reads "TRACKED" instead of the team name. Consider passing the resolved names through (as goalkeeper rows already do).
2. **Forfeits count toward goalkeeper `attemptsFaced`.** A forfeit correctly advances (so it's a kicker official attempt per plan), but since `attemptsFaced` also keys off `advances`, the opposing keeper is credited with an attempt "faced" for a kick that was never taken. Plan §7 only says keeper "attempts faced and saves" without specifying forfeits, and there's no test asserting keeper counts against a forfeit — worth confirming whether a forfeit should be excluded from `attemptsFaced`.
3. **`status` nested-ternary readability.** The abandoned/suspended/sudden-death/next-side branches form a right-associative chain but are indented as if peers of `decided`, which reads misleadingly. Logic is correct; consider an explicit if/else helper.
4. **Attempt-button accessible name** is just the outcome label (kicker+side only via `title`, which screen readers don't reliably announce). Minor — consider an `aria-label` including side/kicker.

### Verdict
No blocking issues. All §9 SOC-6B4 items are covered (conditional visibility, round/early/sudden-death pairing, retake/forfeit handling, anonymous distinctness, participant summaries, isolation from normal totals), tests map cleanly to the list, and the docs correctly close out SOC-6B. The two worth a look are #1 (side-label polish) and #2 (forfeit-vs-attempts-faced semantics).

Rothermal:
Reviewed all four follow-ups against the SOC-4 and SOC-6B contracts and applied them in 523f05b.\n\n- Kicker summaries now show resolved team names instead of raw tracked/opponent values.\n- Forfeited kicks remain official kicker attempts but no longer count as goalkeeper attempts faced. This follows the SOC-4 requirement that the goalkeeper actor remain attached for correction/state consistency without treating the forfeit as a normal save opportunity; the regression test now locks that behavior.\n- Replaced the nested status ternary with an explicit formatter.\n- Attempt buttons now expose side, kicker, and outcome in their accessible names.\n\nVerification: 520 tests pass; lint has 0 errors (3 existing Fast Refresh warnings); production build passes.

Rothermal:
## Follow-up review — `523f05be9 fix: refine soccer shootout review`

The update addresses all four items from my earlier review.

1. **Kicker-table side label** — `KickerRow` now takes a `sideLabel` (`sideName(kicker.teamSide, …)`) and renders the actual team name instead of raw "tracked"/"opponent", consistent with the goalkeeper rows and header. ✔
2. **Forfeit vs. goalkeeper `attemptsFaced`** — now `if (attempt.advances && attempt.outcome !== 'forfeited')`, so a forfeited kick still counts as the kicker's official attempt but no longer credits the opposing keeper with an attempt "faced." A new test locks it in (tracked keeper `attemptsFaced: 0, saves: 0` in the opponent-forfeit scenario); kicker `attempts`/`forfeits` still increment, so the asymmetry is intentional and correct. ✔
3. **`status` nested ternary** — extracted into an explicit `shootoutStatus()` if/else helper; identical logic, now readable. ✔
4. **Attempt-button a11y** — added `aria-label` including side + kicker + outcome, not just the outcome label. ✔

No new issues; changes are minimal, correct, and test-backed. LGTM — this cleanly closes out SOC-6B.

PR #236, opened 2026-07-24
Plan SOC-6C canonical soccer aggregates

Rothermal:
## Summary\n- capture the completed SOC-6C focused Q&A and fixed product contracts\n- define canonical soc_* stat ids, aggregate formulas, result/team/player semantics, and legacy aliases\n- specify migration 047 RPC pagination, RLS, stable-player mapping, completed-publication filtering, and player-merge preservation\n- split delivery into SOC-6C1 engine, SOC-6C2 transport, SOC-6C3 team scopes, and SOC-6C4 player scopes/hardening\n- document route boundaries, observability, benchmark targets, failure states, and verification matrix\n\n## Important discovery\nThe existing merge RPC predates game_participants. Migration 047 must repoint game_participants.source_player_id before deleting a duplicate player, or finalized soccer history becomes unresolved through the FK's ON DELETE SET NULL behavior.\n\n## Verification\n- documentation consistency pass\n- git diff --cached --check\n- no runtime code changed

### DIFF
--- a/docs/AGENT_CODEBASE_OVERVIEW.md
+++ b/docs/AGENT_CODEBASE_OVERVIEW.md
@@ -312,7 +312,7 @@ flowchart LR
 | [`ACCESS_MATRIX.md`](ACCESS_MATRIX.md) / [`PLAN_ADMIN_SECURITY_ROADMAP.md`](PLAN_ADMIN_SECURITY_ROADMAP.md) | SEC-0 through SEC-6 complete; later audit event-family expansion is documented in SEC-6 |
 | [`PLAN_MULTI_GAME_PARKING.md`](PLAN_MULTI_GAME_PARKING.md) | P0–P3b shipped (incl. discard/hydrate race guards); IndexedDB + orphan ops follow-ups remain |
 | [`PLAN_SOC_5_CLOUD_SYNC_AND_FINALIZATION.md`](PLAN_SOC_5_CLOUD_SYNC_AND_FINALIZATION.md) / [`PLAN_SOC_5D_FINALIZATION_AND_RECOVERY.md`](PLAN_SOC_5D_FINALIZATION_AND_RECOVERY.md) | SOC-5 decisions and phases; SOC-5A-D transport through canonical finalization implemented |
-| [`PLAN_SOC_6_SUMMARY_AND_RELEASE.md`](PLAN_SOC_6_SUMMARY_AND_RELEASE.md) / [`PLAN_SOC_6A_SUMMARY_FOUNDATION.md`](PLAN_SOC_6A_SUMMARY_FOUNDATION.md) / [`PLAN_SOC_6B_DETAILED_MATCH_REVIEW.md`](PLAN_SOC_6B_DETAILED_MATCH_REVIEW.md) | SOC-6A Overview and all SOC-6B detailed match-review slices shipped; SOC-6C canonical aggregates are next |
+| [`PLAN_SOC_6_SUMMARY_AND_RELEASE.md`](PLAN_SOC_6_SUMMARY_AND_RELEASE.md) / [`PLAN_SOC_6A_SUMMARY_FOUNDATION.md`](PLAN_SOC_6A_SUMMARY_FOUNDATION.md) / [`PLAN_SOC_6B_DETAILED_MATCH_REVIEW.md`](PLAN_SOC_6B_DETAILED_MATCH_REVIEW.md) / [`PLAN_SOC_6C_CANONICAL_AGGREGATES.md`](PLAN_SOC_6C_CANONICAL_AGGREGATES.md) | SOC-6A Overview and all SOC-6B detailed match-review slices shipped; SOC-6C canonical aggregate Q&A and implementation plan complete |
 | [`PLAN_BASKETBALL_EVENT_MODEL_ROADMAP.md`](PLAN_BASKETBALL_EVENT_MODEL_ROADMAP.md) | Required follow-up: BKE-0 planning after SOC-1; no BKE-1+ implementation before SOC-5 |
 
 ### Held / waiting for feedback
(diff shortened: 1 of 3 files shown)


Rothermal:
## Plan review — SOC-6C Canonical Soccer Aggregates

Planning-only PR (no runtime code). Reviewed `PLAN_SOC_6C_CANONICAL_AGGREGATES.md` against the parent SOC-6 roadmap (§SOC-6C and §6 Aggregate Safety Contract) and verified the two load-bearing technical claims against the codebase.

### Verified the two load-bearing claims
- **The migration-041 merge-RPC / FK discovery is accurate.** Confirmed `game_participants.source_player_id uuid references public.players(id) on delete set null` was introduced in migration **043**, while the latest `merge_players_execute` is in **041** — so it predates `game_participants` and does not repoint the column. A post-finalization player merge would null the link via `ON DELETE SET NULL`, orphaning finalized soccer participants. The proposed fix (repoint duplicate→survivor before delete; no uniqueness conflict, so no user-facing resolution) is sound. Good catch.
- **`migration047.test.ts` fits an established convention** — `migration042–046.test.ts` exist and `readFileSync` the `.sql`, asserting it *contains* expected clauses (grants, RLS predicates, signatures). So the file is feasible as planned (see caveat #2).

### Consistency with the parent contract
Cross-checked every SOC-6 §SOC-6C bullet and the §6 Aggregate Safety Contract — all faithful: `soc_*` in `SportConfig`, never writing `game_stats`, RLS-scoped paginated RPCs, client-side deterministic projector as sole authority, completed-active-publications only, abandoned excluded-but-reviewable, stable-player-only cross-game totals, manager-visible unresolved exclusion counts, shootout game-scoped, sum-before-percentage, per-90 deferred. Internally tight too: pagination (20/50) matches across §5/§8/migration contract; shared-clean-sheet credit consistent §4/§9; malformed-vs-transport-failure distinction well drawn; the multi-participant-resolves-to-one-player duplicate-publication edge is explicitly handled; performance targets deliberately kept out of CI assertions.

### Points worth raising (plan-level, non-blocking)
1. **Merge-fix sequencing / coupling.** The FK repoint is a data-integrity fix to an *existing* RPC, but it's bundled into migration 047 and only ships with SOC-6C2. That's acceptable **only because** soccer is dev-only through SOC-6E (no production finalized soccer data yet) — the plan should state that justification explicitly, and note that any environment already holding a finalized soccer game plus a past merge may *already* have a nulled `source_player_id` that 047 won't retroactively recover. Is a repair/backfill needed, or is a clean dev state assumed?
2. **"Automated" overstates what the test convention can prove for SQL behavior.** Since `migrationNNN.test.ts` only string-matches the `.sql` text (no pgTAP/Postgres in CI), §13's automated "migration 047 … filters, ordering, grants, and role checks" can confirm the SQL *says* the right things but not that it *behaves* — RLS actually blocking inaccessible teams, keyset pagination returning stable rows, and crucially the merge repoint actually preventing the null. Those are the highest-risk items and belong to the manual Supabase matrix (which the plan does include). Worth stating explicitly so the string-contract test isn't mistaken for behavioral proof.
3. **Align C2 acceptance wording with #2.** "player merge preserves finalized soccer source links before deleting the duplicate" reads as a runtime guarantee but is only auto-checkable as a string match; consider flagging it as manual/integration-verified, or add a genuine runtime integration check for that one repoint.

### Verdict
Strong, thorough plan that stays faithful to the SOC-6 roadmap and surfaces a real latent data-integrity bug with a correct fix. No blocking issues — the three items are refinements (make the dev-only justification and the automated-vs-manual verification boundary explicit, especially around the merge repoint).

Rothermal:
Reviewed all three plan-level comments and accepted them in 8ccc836.\n\n- Added the explicit dev-only sequencing justification for coupling the existing merge-RPC fix to SOC-6C2.\n- Removed the clean-database assumption. The plan now requires a pre-047 audit plus a conservative backfill that follows only audited duplicate-to-survivor merge lineage, verifies the current survivor and team membership, and leaves unprovable rows unresolved.\n- Clarified that migration047.test.ts is a static SQL contract test, not PostgreSQL behavioral proof. RLS, equal-timestamp keyset behavior, historical repair, and merge preservation are explicitly manual/integration-verified in Supabase.\n- Updated C2 acceptance and rollout steps to reflect that verification boundary.\n\nNo runtime code changed; git diff --check passes.

Rothermal:
## Follow-up review — `8ccc83627 docs: refine SOC-6C migration verification`

All three points from my earlier review are fully addressed, and the update goes further.

I also confirmed the new repair procedure is technically grounded: `player_merge_audit` (migration 024) has both `duplicate_player_id` and `survivor_player_id` (NOT NULL), so the proposed recursive `duplicate → survivor` lineage walk references real columns. And since the audit table (024) predates both `game_participants` (043) and soccer finalization, it covers every merge that could have nulled a soccer source link — no under-recovery window for the relevant data.

1. **Merge-fix sequencing + already-nulled dev data** — now states the dev-only justification explicitly, and rather than assuming a clean DB, adds a conservative pre-047 repair: find null `source_player_id` rows, walk `player_merge_audit` lineage recursively to the current survivor, restore only when the survivor exists and belongs to the participant's team, never infer from name/jersey/prose, and leave unprovable rows unresolved (reported truthfully). Plus a new `audit_soccer_participant_sources_pre_047.sql` script and SQL-notice repaired/remaining counts. Exactly the right conservative shape, consistent with the plan's "never merge by name/jersey" principle.
2. **"Automated" overstating SQL-behavior proof** — added an explicit note that CI has no Postgres/pgTAP, so `migration047.test.ts` is a static SQL-contract (text-contains) test only; the automated item was reworded to "SQL text contains …", and RLS/cursor/repair/transaction behavior is now clearly assigned to the manual matrix (with two new repair-scenario rows).
3. **C2 acceptance wording** — now split into three explicit tiers (static contract tests / TS tests against mocked RPCs / manual Supabase behavior); the merge item distinguishes "a new merge" vs "an audited historical merge repairs its null link, unprovable stays unresolved"; and a caveat states passing `migration047.test.ts` alone doesn't prove the DB behavior. Rollout §14 reorders to run the pre-047 audit → apply → review notices → rerun → manually verify RLS/pagination/repair/merge.

No new issues; this closes out all three refinements. LGTM — ready to merge.

PR #237, opened 2026-07-24
docs: plan BKE-0 basketball event architecture

Rothermal:
## Summary

Authors the **BKE-0** architecture plan for migrating basketball off its counter / action-log / shot-record model and onto the shared `GameEvent` foundation proven by SOC-1 through SOC-6B. The roadmap has always gated BKE-1+ on "SOC-5 complete and BKE-0 approved" — SOC-5 is complete, and this fills the missing BKE-0.

Docs only. No `src/` or migration changes.

Now at **revision 4**, with three review passes folded in.

## Audit findings that shape the plan

- The shared platform in `src/lib/gameEvents/` needs no envelope, storage, ordering, or projection change — but it does need **one addition**: `update`/`delete`/`restore` are single-event (`mutations.ts:154-210`) while only append batches, so a validated atomic multi-event mutation lands in BKE-1.
- `SportGameState`, `normalizeSportGameState`, and the fingerprint helper live in the soccer module; they must move to a sport-neutral home first. That extraction is BKE-1's first, behavior-preserving step.
- `rebuildGameEventProjection` overwrites stats/scores/shot chart and **clears `actionLog`**, and `gameReducer.ts:120` no-ops all twelve legacy aggregate actions once a stream exists — **all-or-nothing**, not scoped to what a phase has replaced.
- **Every binding, recorder, finalization, and reopen RPC in migrations 043-046 is hard-gated on `sport_id = 'soccer'`**, and `game_event_canonical_publications` pins it by CHECK constraint. The roadmap did not anticipate this, and it is what drives the BKE-4 split.
- `ShotRecord` requires x/y/zone/2pt|3pt, so free throws and unlocated shots have never been in the shot chart — which narrows what Clear Shot Chart is allowed to touch.

## Recommendations recorded

**Nothing in this plan is approved.** Every §11 row — including the clock phasing and the cloud RPC strategy — is a recommendation pending BKE-0 approval.

- Assist is an **actor role on the shot** (matching soccer); rebound is a **separate linked event**; block is a separate event and the **sole** representation (no `blocked_by` actor).
- Decrement is a **per-stat contract**, not one rule: tombstone for standalone events, revise-the-shot for `ast`, confirmed multi-event removal for shots, named dual consequence for fouls.
- Linked-event integrity is explicit: stale links are cleared or tombstoned, never retained invalid; a removed shot takes its linked **rebound** but only **unlinks** its block; grouped delete implies grouped restore.
- Rules are **snapshotted at setup** — for **new event games only**; legacy games keep their current retroactive bonus behavior.
- `min` gets a producing event, `basketball.minutes_adjustment`; decrement appends `-1` rather than tombstoning, since adjustments carry arbitrary signed values.
- `BasketballMatchParticipant` carries **`teamSide`**, because `GameEventActor` has none and soccer's implicit "a `participantId` means tracked side" rule breaks once opponent participants are allowed.
- Compatibility aggregates are **disposable compatibility projections** — derived, rebuildable, projector-versioned, fingerprint-inert, terminal per reader — never a dual write.
- The clock ships as **BKE-5, a fast follower**; the catalog reserves the clock/stoppage/substitution types up front so BKE-5 needs no envelope, table, or ordering change.

## Phasing and rollout

- **BKE-4 splits into 4A-4D**: 4A neutral RPC extraction with soccer-parity exit, 4B basketball transport, 4C recorders and finalization, 4D aggregates and cutover. As one phase it mixed a change that must alter nothing observable with changes that deliberately alter a lot.
- **Two gates, not one.** Event-game creation stays behind an **internal gate through BKE-3**; the user-visible per-game opt-in ships with **BKE-4D**, once capture and cloud lifecycle are both complete. A BKE-1..3 event game is intentionally incomplete and must not reach a user.

## Decisions requested at approval (§12a)

Two authority decisions are requested **now**, not deferred — they determine what the projector and cloud layer may treat as truth, so BKE-1 needs them settled. Only their mechanics belong to later phases.

1. **Finalized corrections.** Proposed: event games adopt soccer's reasoned reopen and republication and retire `stat_corrections`; legacy games keep it unchanged. Delivered in BKE-4C.
2. **Aggregate authority.** Proposed: canonical publications become the durable authority the way SOC-6C makes them for soccer; legacy tables get compatibility rows only until each reader migrates. Delivered in BKE-4D, whose scope this answer sizes.

§12b holds five confirm-only questions: F12 out-of-order removal, free-throw grouping, opponent individual tracking, setup friction, and shot-ordinal stability/population.

## Also updated

- Roadmap: status, phase table with the 4A-4D split and gating note, recommendation labels, regression requirements.
- Agent overview: active-work row.
- F13 hold note: the event model **supersedes** its `ShotRecord` extension and `shot_chart` columns; product intent is preserved and delivered in BKE-3. The shot ordinal stays off `ShotRecord`, in a separate review projection.
- `.gitignore`: ignore the local `.vite` cache directory.

## Verification

`pnpm lint` (0 errors, 3 pre-existing warnings), `pnpm test` (520 passing), `pnpm build` — all clean at the original commit; every revision since has been documentation-only. Merged `origin/stattracker` in and resolved the overview-table conflict with PR #236 (SOC-6C).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Rothermal:
Overall, this is a strong architecture audit and the right long-term direction. The event authority boundary, no-backfill rule, immutable setup snapshot, and F13 reconciliation all make sense. I do think the draft needs another revision before BKE-0 is approved, mainly around phase safety and compound-event correction semantics.

## Blocking concerns

1. **BKE-1 cannot safely be a user-visible opt-in phase as currently scoped.** The moment `eventStream` is non-null, the reducer disables the full legacy stat grid, score controls, period control, clear/undo actions, and other aggregate mutations. BKE-1 only replaces court shot/assist/rebound capture; BKE-2 replaces the rest, and cloud persistence does not arrive until BKE-4. That leaves a BKE-1 event game incomplete and local-only. I recommend keeping event-game creation behind a development/internal gate through BKE-3, then enabling the per-game user opt-in only once BKE-4 supplies a complete capture and cloud lifecycle. If an earlier production rollout is intended, the phase boundaries need to move so every enabled game remains fully usable.

2. **Manual minutes are claimed as covered but have no producing event.** `min` remains writable under `clockModel: 'none'`, while `INCREMENT_STAT`/`DECREMENT_STAT` become no-ops and the catalog has no minutes event. Please define a `basketball.minutes_adjustment` (or narrowly constrained manual-stat event) with explicit increment/decrement/edit behavior. BKE-5 can then retire or ignore those events only for anchored-clock games without changing historical clock-less games.

3. **“Decrement tombstones the newest event matching participant + stat” is unsafe for compound events.** Decrementing `ast` cannot tombstone its made-shot parent without also deleting the basket and attempt. Likewise, a shot can project score plus shooting totals, and a foul may project both player and team totals. Please replace the blanket rule with a stat-to-mutation contract: standalone scalar event -> tombstone; assist -> revise/remove the assist actor from the shot; rebound/block/turnover -> tombstone the linked standalone event; compound foul/shot -> use a domain edit/remove flow with an explicit user consequence.

4. **Linked-event integrity and multi-event edits need a contract.** F13 can change a made shot to missed, change the shooter/side, or remove a shot that has a rebound/block relationship. The shared engine can atomically append a batch, but update/delete currently operate one event at a time. “Advisory links” are fine for projection, but not enough for a detail editor that promises durable relationships. Please define whether stale links are rejected, automatically revised/tombstoned, or intentionally retained, and add a validated atomic multi-event mutation path if one user edit can affect several rows. Also choose one block representation: `blocked_by` on `basketball.shot` and a separate `basketball.block` event currently duplicate the same fact and can drift.

5. **The existing Clear Shot Chart workflow is not accounted for.** Today `clearEntireShotChart` removes all shots and reverses linked assist/rebound stats even when unrelated actions trail them. In the event model this is a multi-event correction, not simply clearing a projection. Please state whether the command is retained (batch tombstone shots plus dependent relationships), changed to scrub only locations/chart visibility, or retired with an intentional UX replacement. It belongs in BKE-1 acceptance/regression scope either way.

## Model and F13 concerns

6. **Envelope `sequence` is not a durable shot number.** It is recorder capture order across every event, so fouls/rebounds/timeouts interleave; historical additions and multiple recorder streams make the gap larger. F13 should derive an effective shot ordinal from ordered active `basketball.shot` events, or define an immutable shot-specific ordinal. It should not display raw event sequence as “shot number.”

7. **Tighten shot/link validation.** I recommend deriving `located` from `location !== null` instead of storing both; requiring free throws to be value 1 with no court location; requiring field goals to be value 2 or 3; and defining zone/location consistency. Relationship validation should require assists on made same-side shots (not the shooter), rebounds linked only to missed attempts with kind checked against side, blocks linked to eligible missed field goals, and all links to remain within the same game.

8. **Capture preferences are resumable UI state, not projection truth.** Persisting selected participant/side is useful, but the projector should not need them “to rebuild.” Please explicitly exclude capture preferences from canonical publication/final aggregate semantics and clarify whether changing them should dirty cloud sync/fingerprints.

9. **Team pseudo-player treatment is contradictory.** Section 4.1 says they are match participants, while sections 11/12 leave that open. My recommendation is team-kind actors plus projection into the existing `__team_home__` / `__team_opp__` rows, without putting pseudo teams into the player participant roster; that avoids contaminating future player/lineup summaries. Whichever answer we choose should be stated once as a proposal until Q&A settles it.

## Phase and cloud recommendations

10. **BKE-4 is too broad for one implementation phase/PR.** It combines a behavior-preserving generalization of a large, proven soccer SQL surface with basketball binding, sync/recovery, recorder resolution, finalization, corrections, compatibility aggregates, and rollout. For reviewability and context-window safety, I suggest BKE-4A neutral RPC extraction with soccer parity tests, BKE-4B basketball bind/sync/recovery, BKE-4C recorder/finalization/reopen, and BKE-4D aggregate compatibility plus production cutover.

11. **Avoid describing compatibility aggregates as “dual projection” or an indefinite dual write.** Events should remain the only authority. If `game_stats`/`shot_chart` are temporarily populated, call them disposable compatibility projections and define their rebuild/version/fingerprint behavior. My recommendation for the two load-bearing open questions is: event games use reasoned reopen/republication rather than a `stat_corrections` overlay, and canonical publications become the durable aggregate authority; legacy tables may receive temporary derived compatibility rows only until each reader migrates.

## Recommended answers for the remaining Q&A

- Keep F12 strict newest-first through BKE-2; deliver arbitrary removal with the BKE-3 timeline/detail UI.
- Store one event per free throw, with an optional shared `tripId`/sequence metadata later if 1-and-1 or trip review needs grouping. A grouping event is unnecessary for totals.
- Allow opponent participants in the schema now, but do not require opponent roster setup; team/unknown actors remain valid.
- Snapshot resolved rules and roster while preserving the current one-tap setup path where defaults are sufficient.

Two wording cleanups: the document is explicitly a draft, but cloud generalization and BKE-5 are labeled “approved”; please cite where those were settled or relabel them as recommendations pending this Q&A. Also clarify that the immutable rules snapshot fixes retroactive bonus behavior for new event games only; legacy aggregate games retain their historical behavior unless separately addressed.

With these points folded in, I think this becomes a solid base for the phase-specific Q&A and plans.

Rothermal:
Re-review complete. This revision is substantially stronger and addresses nearly all of the first-pass feedback well: the internal/user gate split, BKE-4A-D breakdown, per-stat decrement contract, atomic mutation requirement, link validation, pseudo-team decision, compatibility projection contract, and F13 ordinal correction are all good improvements.

I found three remaining model issues that should be corrected before approval, followed by a few smaller consistency questions.

## Remaining substantive issues

1. **Clear Shot Chart now has a broader destructive scope than the current command.** Section 6.3 says to tombstone *every* active `basketball.shot`. That includes free throws and unlocated quick-entry shots, neither of which can exist in the current `ShotRecord[]` (`ShotRecord` requires x/y, zone, and 2pt/3pt) and neither is removed by today's `clearEntireShotChart`, which starts from `state.shotChart`. The event command should target only shots represented in the projected chart, likely located field-goal events, plus the linked dependents whose current behavior is intentionally preserved. Otherwise “Clear Shot Chart” silently erases unrelated FT and unlocated scoring stats. The equivalence fixture should explicitly include FT and unlocated attempts and prove they survive.

2. **The manual-minutes decrement arithmetic is incorrect for valid edited/signed histories.** Sections 4.2 and 6.1 say decrement tombstones the newest adjustment, but an adjustment may be any signed non-zero integer. Tombstoning `+5` reduces minutes by five, and tombstoning a newest `-1` increases the total. The cleanest contract is to append `deltaMinutes: -1` whenever projected minutes are at least one; alternatively revise the newest positive adjustment by exactly one and tombstone it only when it reaches zero. Every add/update must validate the resulting participant total stays non-negative. I would also avoid flooring the projection at zero: flooring hides an invalid stream instead of diagnosing it, so a negative sum should make the stream incomplete until repaired.

3. **Opponent participants need a side in their stable match identity.** Section 4.1 defines participant id/player id/name/number/status, while §4.5 validates that assist and linked actors are on the same or opposite side and §12 explicitly allows future opponent participants. `GameEventActor` itself has no side; only the event envelope does. Add `teamSide: 'tracked' | 'opponent'` (or equivalent) to `BasketballMatchParticipant`, then validate each player actor against both the participant roster and the event side. Without it, same-side assist validation and opponent-participant attribution cannot be deterministic.

## Consistency and lifecycle questions

4. **Block deletion behavior contradicts its catalog rationale.** Section 4.2 says a block is separate partly because it “needs to survive its shot being edited or removed,” while §6.2 and §6.3 tombstone linked blocks when a shot is removed. Choose one rule. My preference: an ordinary shot edit/removal presents a choice or named consequence; Clear Shot Chart should only remove relationships/stat events that match today’s behavior, which currently includes linked assist/rebound but not a linked block feature. Also define restore symmetry: when a shot and dependents were tombstoned atomically, does restoring the shot offer/perform an atomic restore of those dependents? Otherwise delete is grouped but undo/restore is not.

5. **Projected shot ordinal should not require a new durable `ShotRecord` field by accident.** Section 5 says the projector “carries it on the projected `ShotRecord`,” but §8 says the proposed `ShotRecord` extension is superseded and the current type has no ordinal. Prefer deriving the display ordinal from the index of the ordered active shot-event/detail view, keyed by event id. If a transient field is actually required, name a separate basketball shot-review projection type so it is not mistaken for persisted `shot_chart` metadata.

6. **The timing of open questions 1 and 2 is contradictory.** §12 calls both load-bearing before BKE-1, then says question 1 is decided by BKE-4C and question 2 before BKE-4D. Either settle them during BKE-0 and remove the later-decision wording, or explicitly defer them and stop presenting them as BKE-1 gates. I lean toward settling the authority model now, while leaving implementation detail to 4C/4D.

Two minor wording fixes:

- Capture preferences cannot support “device handoff” if they are intentionally excluded from cloud sync/publication; they support reload and local park/resume. A second device should initialize its own preferences.
- A team-kind `GameEventActor` currently requires a `label`; §4.5’s “carrying only a side” should either mention that label or identify a planned generic actor-type change.

After those are reconciled, I would consider the architecture ready for its focused decision Q&A. CI is green, and I found no concerns with the other documentation changes.

Rothermal:
Changes were required — all six points held up against the code, and three of them were real defects rather than wording. Pushed as `7af1e84`.

## Substantive issues

**1. Clear Shot Chart scope — confirmed, and worse than a wording problem.** `ShotRecord` (`src/types.ts:11-20`) requires `x`, `y`, `zone`, and `shotType: '2pt' | '3pt'`, and `clearEntireShotChart` starts from `state.shotChart`, so free throws and unlocated baskets have never been reachable by the command. "Tombstone every active `basketball.shot`" would have deleted FT and unlocated scoring stats that the button leaves alone today.

§6.3 now scopes the command to active `basketball.shot` events with `attempt: 'field_goal'` **and** `location !== null`, plus their linked rebounds, with an explicit Preserved list. Added your fixture requirement to §6.3 and §13: the BKE-1 equivalence suite must contain made and missed free throws and at least one unlocated field goal, and assert they are unchanged after a clear. Without that case the regression is invisible, since every located-only fixture passes either way.

**2. Minutes arithmetic — confirmed.** Tombstoning a `+5` drops five minutes for one tap, and tombstoning a newest `-1` raises the total. §4.2 and §6.1 now use your primary contract: decrement appends `deltaMinutes: -1` whenever projected minutes are ≥ 1 and is disabled otherwise, never tombstoning. Every add and update validates a non-negative resulting total.

I took the flooring point too. The projection is no longer floored: a negative sum is only reachable through a tombstone or a merged multi-recorder stream that validation never saw, so flooring would hide exactly the case worth surfacing. A negative total now makes the stream **incomplete** under §5 — displayable, diagnosable, not finalizable.

One carve-out I made explicit: F12 newest-first **undo** still tombstones the newest adjustment. That is undoing a specific recorded action rather than doing arithmetic on a total, and it is correct precisely because it targets an event the user just created.

**3. Participant side — confirmed.** `GameEventActor` carries no side (`gameEvents/types.ts:12-33`); only the envelope does. Soccer gets away with it because its roster is tracked-team-only, so `validateActorSide` (`soccer/projector.ts:849-876`) can read "has a `participantId`" as "is on the tracked side" — an inference that dies the moment §12 allows opponent participants. `BasketballMatchParticipant` now carries a required `teamSide`, and §4.5 gained an actor/side validation table checking each player actor against both the roster and the envelope. §4.1 states why basketball is explicit where soccer is implicit.

## Consistency and lifecycle

**4. Block deletion — you were right, and today's code settles it.** `LINKED_COURT_STAT_IDS` is `['ast', 'oreb', 'dreb']` (`clearShotChart.ts:14`) with no block, so removing linked blocks would have invented behavior, not preserved it. Resolved toward the catalog rationale: a linked block is **unlinked, not removed**, when its shot is removed or edited — it stays a standalone defensive stat. A linked rebound is still tombstoned with the shot, since it describes that specific miss and today's command already reverses it. §6.2 states the rebound/block distinction as a rule rather than leaving it implied.

Restore symmetry is now defined. Immediate undo of a grouped removal restores the whole group atomically. A later timeline restore restores the shot and then offers to restore each tombstoned dependent still pointing at it — the tombstoned rebound retains its `relatedEventId`, so the group is recoverable from the stream with no group-id column and no engine change. Unlinked dependents are never silently re-linked; re-linking is an explicit edit.

**5. Shot ordinal — confirmed contradiction, fixed.** §5 no longer puts the ordinal on `ShotRecord`. It is derived in a separate `BasketballShotReviewEntry { eventId, ordinal, … }` owned by the review/detail layer, so the persisted type and the `shot_chart` mapping in `cloudSync.ts` are untouched and §8's "the `ShotRecord` extension is superseded" stays true.

**6. Open-question timing — fixed, in the direction you leaned.** §12 now splits: **§12a** holds the two authority decisions requested at BKE-0 approval, since they determine what the projector and cloud layer may treat as truth and BKE-1 needs that before writing projector code; **§12b** holds the remaining confirm-only questions. BKE-4C and BKE-4D are named as *delivering* the mechanics, not deciding them. The "load-bearing before BKE-1, then decided by 4C/4D" contradiction is gone.

## Minor wording

Both taken. Capture preferences now say reload and local park/resume, and state that a second device initializes its own — the handoff claim was incompatible with excluding them from sync. §4.5 notes that a `team`-kind actor is a `GameEventLabelActor` with a **required `label`** and no `participantId`, with side coming from the envelope; no actor-type change is proposed. Basketball populates `label` from the snapshotted team designation so a published stream reads without resolving the roster.

## Nothing deferred

All six points and both wording fixes are in this revision. §11 gained rows for participant identity, restore symmetry, and the narrowed clear-chart scope; §13 gained regression rows for FT/unlocated survival, grouped restore, minutes arithmetic with signed histories, and actor/side validation.

Docs-only change, so CI coverage is unchanged from the previous green run.

Rothermal:
Third review complete. The latest revision resolves the substantive findings from both prior passes. I found no remaining blocking architecture issue: clear-chart scope now matches `ShotRecord[]`, signed minutes behave correctly, participant-side attribution is explicit, blocks/rebounds have distinct lifecycle rules, grouped restore is defined, shot ordinals stay out of persisted `ShotRecord`, and the authority decisions are requested at the correct point. CI is green.

I have two cleanup items before merge, plus one product clarification that can live in the focused Q&A:

1. **The clear-chart equivalence assertion is phrased impossibly.** Section 6.3 says the preserved FT/unlocated events “and their stats” should be byte-identical to the pre-clear projection. Their event rows can be byte-identical, but aggregate `2pt`/`3pt` totals cannot be: located and unlocated field goals share those same stat ids, so clearing located attempts must change the combined counters and score. The fixture should instead assert:
   - FT and unlocated event envelopes/revisions are byte-identical;
   - post-clear counters equal the expected contribution of the preserved events after subtracting charted attempts; and
   - free-throw totals are unchanged while shared field-goal totals change only by the removed charted contributions.

2. **The PR description has drifted behind the document.** It still says decrements “tombstone the newest matching event,” calls BKE-4/BKE-5 decisions settled, and describes the original open-question state. Please update the PR body to summarize the per-stat decrement contract, recommendations-pending-approval status, BKE-4A-D split, internal gate through BKE-3, and §12a/§12b decision structure. The code/docs diff is correct; this is review-context cleanup.

3. **Confirm what the displayed shot ordinal counts.** Section 5 currently numbers every active `basketball.shot`, including free throws and unlocated attempts, while the original F13 interaction begins by clicking a charted field-goal marker. Either interpretation is defensible, but it changes what “Shot 8” means to the user. Add this to the stable-label Q&A or state explicitly that the ordinal counts all attempts, not only chart markers. My recommendation is all attempts because they share one authoritative shot event family, with the detail UI making attempt type obvious.

One tiny wording inconsistency: §4.5 says the pseudo-team proposal remains open in §12b, while §12 says that question is closed. It should say “pending BKE-0 approval” rather than pointing to §12b.

After those doc cleanups, this is ready for the focused BKE-0 decision Q&A. No further architecture rework is needed from my review.

Rothermal:
All four items addressed — doc fixes in `dee73f9`, and the PR description is rewritten.

**1. Clear-chart assertion was phrased impossibly — you're right.** Located and unlocated field goals share the `2pt` / `3pt` / `*_miss` ids, so clearing charted attempts necessarily moves those counters; "preserved events *and their stats* byte-identical" could never hold. §6.3 now states the three separable assertions you listed:

| Assertion | Scope |
|---|---|
| Preserved FT and unlocated event envelopes/revisions byte-identical | The stream |
| `ft` / `ft_miss` unchanged | Ids the cleared events cannot contribute to |
| `2pt` / `3pt` / `*_miss` and score down by exactly the charted attempts' contribution | Ids shared between charted and non-charted shots |

I added why all three are needed rather than just the first: the row assertion catches a command whose **scope** is too wide, the counter assertions catch one with the right scope but the wrong **arithmetic** — reversing a preserved event's contribution, or double-subtracting a shared one. Those fail independently. §13's regression row was reworded to match.

**2. Shot ordinal population — taking your recommendation, all attempts.** §5 now states it explicitly rather than leaving it inferable: the ordinal counts every active `basketball.shot`, free throws and unlocated field goals included. One authoritative event family, and a chart-only count would make the same shot "Shot 8" on the chart and "Shot 12" in the timeline.

I also wrote down the cost you implied, since it's a real UI obligation rather than a footnote: entering the detail view by tapping a chart marker can land on a non-contiguous number, so the detail surface must show attempt type (FG vs FT, located vs not) beside the ordinal. §12b q5 now carries both confirmations together — stability *and* population — since answering one without the other leaves "Shot 8" ambiguous.

**3. Pseudo-team pointer fixed.** §4.5 now says the proposal is stated once **pending BKE-0 approval** and is no longer carried as an open question in §12b.

**4. PR description rewritten.** It now matches revision 4: the per-stat decrement contract, recommendations-pending-approval framing with nothing marked settled, the BKE-4A-D split, the two-gate rollout with the internal gate through BKE-3, and the §12a/§12b structure with the two authority decisions stated as requested-at-approval. Also folded in the audit findings that emerged during review — the single-event mutation limit, the all-or-nothing reducer gate, and the `ShotRecord` shape constraint that narrows clear-chart.

No architecture changed in this pass; all three doc edits are corrections to how existing decisions were expressed. Ready for the focused BKE-0 decision Q&A, where §12a's two authority questions are the ones that still need your call.

Rothermal:
Final re-review complete at `dee73f9`.

No findings. The last four items are resolved correctly:

- Clear-chart regression assertions now separate preserved event identity from expected aggregate arithmetic.
- Shot ordinal population explicitly includes all attempts, with the chart-entry UX consequence documented.
- The stale pseudo-team §12b pointer is corrected.
- The PR description now accurately reflects the per-stat mutation model, BKE-4A-D split, internal rollout gate, and recommendations-pending-approval status.

I also rechecked the surrounding event catalog, projection contract, correction/link lifecycle, historical authority boundary, phase exits, and regression map. They remain internally consistent after the edits. CI is green.

**Review status: architecture review passes.** This is ready for the focused BKE-0 decision Q&A. The review does not itself approve the §12a authority choices or §12b product confirmations; those remain the explicit decisions for that session, as intended.

PR #238, opened 2026-07-25
SOC-6C1: Add canonical soccer aggregate engine

Rothermal:
## Summary
- replace the legacy soccer SportConfig ids with the canonical soc_* catalog and conservative read aliases
- add pure canonical-publication projection and stable-player cross-match aggregation with explicit quality/exclusion handling
- share one canonical participant-stat mapper between the live projector and aggregate engine
- cover adjusted scores, shared clean sheets, DNP, merged/unresolved identity, abandoned/malformed/ineligible/shootout sources, and duplicate publications
- update the SOC-6C plan, roadmap, agent guidance, and regression matrix

## Verification
- pnpm test (81 files, 532 tests)
- pnpm build


Rothermal:
## Code review — SOC-6C1 canonical aggregate engine

Reviewed against `PLAN_SOC_6C_CANONICAL_AGGREGATES.md` §4, §9, §12, and the existing soccer projector/summary code.

### What this does

Replaces the legacy soccer `SportConfig` ids with the canonical `soc_*` catalog and adds two pure modules: `aggregateStats.ts` (catalog, aliases, formatting, rates, ordering) and `aggregateProjection.ts` (rebuild a canonical publication → match read model → cross-match aggregation by stable cloud player id). No transport, no React, no SQL — correctly scoped to C1, and it stays independently buildable as §14 requires.

The design is sound and the quality/exclusion modelling is genuinely careful: `ok`/`exclusion` result union, per-match unresolved participants with contribution counts, `canManage` retained on every exclusion for later manager-only diagnostics, and `try`/`catch` around the rebuild so one malformed publication cannot take down a scope. Defensive parsing in `snapshotPlayers` is right for untrusted snapshot content.

---

### Findings

**1. The canonical `soc_*` mapping now exists twice, and the two copies already disagree in one case.** *(highest-value item)*

`projector.ts:1182-1236` (`buildProjection`) already maps `SoccerParticipantStatTotals` → the canonical `soc_*` ids onto `Player.stats`. `statsFromReviewRow` (`aggregateProjection.ts:1257-1294`) maps the *same* source shape to the *same* ids a second time, independently. `row.stats` is `SoccerParticipantStatTotals` — identical to the projector's `attacking` — so this is a duplicated derivation of the vocabulary this PR exists to make canonical.

Two formulas are now written twice (`soc_ast = primary + secondary`, `soc_chance_created = keyPasses + primaryAssists`), and one already differs:

| id | `buildProjection` | `statsFromReviewRow` |
|---|---|---|
| `soc_start` | `participant.started ? 1 : 0` | `appearances > 0 && lineupStatus === 'starter'` |

`lineupStatus === 'starter'` is `appearances !== 0 && started` (`summaryPlayers.ts:220-224`), so the two agree only while `started` implies `appearances > 0` in the projector. Nothing enforces that invariant, and if it ever breaks, `Player.stats.soc_start` and the aggregate disagree with no test to catch it. `soc_min_sec` carries the same shape of risk (`totalActiveMs` plus in-flight vs `row.minutesMs`).

Suggested fix: extract one exported mapper (`soccerCanonicalStatsFromTotals(totals)`) and have both `buildProjection` and `statsFromReviewRow` call it, layering only the read-model extras (`soc_app` clamp, `soc_cs`) on top. Note the aggregate genuinely cannot just read `projection.playerStatsById` instead — that map is keyed by local player id and so is empty for exactly the unresolved participants the exclusion reporting needs — so a shared mapper rather than a shared consumer is the fix. Failing that, at minimum add a test asserting the two derivations agree for a healthy fixture.

**2. The legacy alias map has no consumer, and the surface that needs it now reads the wrong ids.**

`SOCCER_LEGACY_STAT_ALIASES` is only consumed by `normalizeSoccerAggregateStats`, which nothing in `src/` calls — the aggregate engine builds stats from review rows, not from stored records. Meanwhile `Leaderboard.tsx:255` iterates `sport.categories` for **any** sport with no soccer branch, so after this merge the soccer Leaderboard offers the 33 canonical ids as sort keys while any existing dev `game_stats` rows still hold `s_goal` / `sot` / `s_tackle`. Those sorts silently read 0 until SOC-6C3 lands.

Soccer is development-only, so this is a transitional gap rather than a production bug — but it deserves an explicit decision: either wire `normalizeSoccerAggregateStats` into the legacy read path now, or record the C1→C3 gap in the plan so it is not mistaken for a regression during manual QA. As it stands, §12's "add the narrow development compatibility map" reads as satisfied when nothing actually reads through it.

**3. `ineligible_source` is excluded silently while quality stays `complete`.**

`aggregateProjection.ts:1157-1164` computes quality from unresolved + malformed + duplicate only. A source failing the `status !== 'final' || cloudScope !== 'team' || !teamId` guard is dropped with `quality: 'complete'`. Since the C2 RPC is supposed to make that state unreachable, hitting it means the server contract was violated — precisely when the plan's "never silently skip malformed canonical data" should apply. `abandoned_match` staying out of quality is correct and intended; `ineligible_source` looks like an oversight.

**4. A `teamId`-less match is counted but not aggregated, and `matchGameRow` asserts past it.**

`aggregateSoccerMatches` is exported and called directly (the tests do), and it does not re-check eligibility. For a match with `game.teamId === null`, line 1088 `continue`s before players/teams accumulate — but the match still lands in `included`, inflating `includedMatchCount` and `metrics.eventCount`, and `matchGameRow` emits `teamId: match.game.teamId!` (`:1401`), putting a `null` into a field typed `string`. Either re-validate at the top of `aggregateSoccerMatches` and push an `ineligible_source` exclusion, or drop such matches from `included` entirely. That non-null assertion is the only one in the file and it rests on an invariant this function never checks.

**5. Duration format deviates from the plan's stated format.**

Plan §4: "Display it as `MM:SS` below one hour." `formatSoccerAggregateDuration` emits `0:00` and `59:59` (asserted in `aggregateStats.test.ts`), i.e. `M:SS`. Harmless either way, but pick one — pad to `00:00` / `05:03`, or update the plan wording to `M:SS`.

**6. Smaller items**

- `addSoccerAggregateStats` mutates its first argument while every sibling helper is pure. It is used correctly (targets are always clones), but the name reads as pure. Consider `addSoccerAggregateStatsInPlace`, or return a new object.
- `existing.rates = soccerAggregateRates(existing.stats)` (`:1118`) runs per player per match and is then discarded by the recompute in the final `.map` (`:1173`). Dead work on the hot path.
- `SoccerAggregatePlayer.matchIds` includes matches where the player was a DNP roster entry, so `matchIds.length` is *not* games played (`soc_app` is, per §4). Worth a doc comment before a destination page reaches for the convenient-looking field.
- `mergeMatchPlayers` caps `soc_app` / `soc_start` / `soc_cs` with `Math.max` but **sums** `soc_min_sec` across participants resolving to one player. Correct for two separate stints; double-counts if two participant rows can ever cover overlapping intervals. Worth confirming that is impossible.
- `exclusion.message` carries the raw `error.message` from the rebuild. §4 restricts detailed projection diagnostics to owner/admin — `canManage` is present, so this is fine, but C3/C4 must gate on it rather than rendering `message` for everyone.
- `dates.sort()` is lexicographic, correct only if `game.date` is ISO `YYYY-MM-DD`. It is typed as a bare `string`.

---

### Test coverage

Good breadth for a pure-engine slice — adjusted scores, shared clean sheets, DNP, merged identity, unresolved separation, abandoned/malformed, shootout exclusion, and exact-vs-conflicting duplicates. The shootout case pins a 0-0 shootout as `draws: 1`, which confirms the score-derived W/D/L is deliberate.

Gaps worth closing:

- no test asserting the aggregate's stats equal the projector's `Player.stats` for the same fixture (finding 1);
- no `ineligible_source` case (finding 3) and no `teamId`-null case (finding 4);
- no coverage of `normalizeSoccerAggregateStats` against real stored data, because nothing calls it (finding 2).

One documentation note rather than a defect: the engine derives W/D/L purely from `sideTotals`, so a knockout tie decided on penalties records as a **draw** even though `projection.result` is `tracked_win` / `opponent_win` and `decidedStage` is `shootout` (`projector.ts:1001-1016`). That matches league convention and §9's "match result uses final canonical score", and it is tested — but Tournament Stats in C3 will surface it, and the plan never states it. Worth one line in §9 so it is not re-litigated later.

### Security and performance

No new network, SQL, or auth surface; RLS remains a C2 concern. An isolated `createInitialState()` per publication is the right call — no shared mutable state between sources. `structuredClone` on snapshot-derived data is appropriate. Performance is linear in matches × participants with a 33-key object per participant; the only waste is the redundant rate recompute above. The §12 50-match / 10k-event fixture correctly belongs to C2.

### Verdict

Sound, well-scoped work. Finding 1 is the one I would fix before merge — it undercuts the "one canonical vocabulary" goal that motivates the slice, and it is cheap to fix. Findings 2-4 are small and self-contained; 5-6 are polish.

Rothermal:
Reviewed each finding against the current projector and SOC-6C boundaries. Follow-up commit: 5e1a0e8.

Implemented:
- Extracted soccerCanonicalStatsFromTotals and now use it in both buildProjection and aggregate projection. A healthy fixture asserts exact projector/aggregate stat parity.
- Ineligible sources now make aggregate quality partial. Both canonical-source projection and direct aggregateSoccerMatches input validate final team scope plus a real ISO date.
- Direct null-team matches are excluded before counts/metrics/game rows; the unsafe non-null assertion is gone.
- Duration output now follows the plan exactly: padded MM:SS below one hour.
- Renamed the mutating accumulator, removed redundant per-match rate recalculation, documented matchIds as source membership rather than games played, and documented why merged participant time/event contributions are preserved.
- Documented that penalty-decided normal ties remain draws in aggregate W-D-L.

Deliberately deferred/clarified:
- The legacy alias map remains an exported/tested C1 primitive and is not wired into the generic legacy Leaderboard. C1 intentionally has no route integration, and canonical soccer destinations arrive in C3. The plan now explicitly says C3 must not mix legacy game_stats with canonical publications; any temporary pre-release fallback must opt into the alias map and identify itself as non-canonical.
- Raw exclusion messages still carry canManage for C3/C4 presentation gating; there is no aggregate UI in C1 that can expose them. The existing manager-only contract remains unchanged.

Verification: 81 test files / 532 tests passed, lint 0 errors (3 existing Fast Refresh warnings), and production build passed.

Rothermal:
## Re-review — `5e1a0e8`

Good follow-up. Every finding is addressed, and the two structural ones are fixed properly rather than papered over: `soccerCanonicalStatsFromTotals` is a real single source for the vocabulary, and `EligibleSoccerAggregateMatch` with the `isEligibleAggregateMatch` type predicate removes the non-null assertion by making the invariant checkable instead of assumed. The deferral on the alias map is reasonable and the plan now records the C3 constraint, which was the point.

I verified the two behavior changes the refactor introduced into the shipped projector:

- `soc_app` moved from `+= participant.appearances` to `+= (appearances > 0 ? 1 : 0)`. Safe — `appearances` is only ever assigned `1` (`projector.ts:281`), `Math.max(1, …)` (`:481`), or `0|1` at creation (`:556`), so it can never exceed 1 and the clamp is a no-op.
- `soc_start` moved from `started` to `appeared && started`. This is the divergence from my finding 1, resolved by adopting the stricter rule into the projector. Correct direction.

One item survives, and it is the same root cause in two places.

---

### 1. The projector now authors `soc_cs`, and the parity test can't see it

`buildProjection` passes `cleanSheet: false` unconditionally (`projector.ts:1207`), then writes every id in `SOCCER_CANONICAL_STAT_IDS` — so `Player.stats.soc_cs` is now materialized as an explicit `0` for every soccer participant. Previously the key was never written at all.

That contradicts plan §4: "`soc_cs` is a read-model stat derived from the reviewed clean-sheet eligibility rules. It is not a client-authored event total." The projector cannot compute it (it has no SOC-6B eligibility context, hence the hardcoded `false`), so it now emits a permanently-wrong zero for an id it is not supposed to own. Nothing consumes it today — soccer aggregate cloud sync is disabled once sport-owned setup exists, so these never become `game_stats` rows — but if soccer sync is ever enabled, `soc_cs: 0` ships as client-authored data, which is precisely the case §4 was written to prevent.

**The parity test does not catch this, and reads stronger than it is.** The new assertion:

```ts
expect(striker.stats).toEqual(
  fixture.state.players.find(player => player.id === 'striker-local')?.stats
)
```

passes only because `cloud-striker` has `soc_cs: 0` on both sides. The *same fixture* has `cloud-keeper-a` and `cloud-keeper-b` at `soc_cs: 1` (the shared clean sheet the test name advertises) — the aggregate gives 1, the projector gives 0, so an identical `toEqual` on either goalkeeper fails today. So "a healthy fixture asserts exact projector/aggregate stat parity" holds only for the one player where the single divergent id happens to be zero, and a future regression in any goalkeeping-only field would sail through.

Suggested fix, roughly two lines plus a test:

- keep `soc_cs` out of the projector write — either have `buildProjection` skip it in its accumulation loop, or give `soccerCanonicalStatsFromTotals` a participation shape where `cleanSheet` is `boolean | null` and `null` means "not authored here";
- extend the parity assertion to `cloud-keeper-a` over the remaining ids, so the guarantee covers a goalkeeper and `soc_cs` is documented as the one deliberate exception rather than an accident of which player was sampled.

---

### 2. Forward-looking note for C2: `isIsoDate` is now a hard eligibility gate

Stricter than what I flagged — I only noted that `dates.sort()` assumed ISO. Making it an eligibility predicate is the better call, and I verified it is safe against current data: `GameSetup.tsx:60` seeds `new Date().toISOString().split('T')[0]`, and `cloudSync.ts:981` maps `game_date` straight through, so `YYYY-MM-DD` holds end to end.

The risk is at the C2 boundary. If migration 047 returns `game.date` from a `timestamptz`, or as a full ISO datetime, `isIsoDate` rejects **every** source and the result is an empty aggregate reported as `partial` rather than a loud failure — a confusing thing to debug at integration time. Worth pinning the exact `YYYY-MM-DD` shape in the §5 RPC item contract next to `game: { id, date, status, … }`.

### 3. Noted, no action

Participants with no `participantStats` entry now receive all 33 ids as explicit zeros instead of only `soc_app`/`soc_start`/`soc_min_sec` (the old `continue` path). Verified this cannot reach `game_stats` for the reason above; mentioning it only so the sparse→dense change in `Player.stats` is a known consequence rather than a surprise later.

The new ineligible-source test is well targeted — `2026-02-30` exercises the round-trip check rather than just the regex, both entry points are covered, and asserting `metrics.eventCount === 0` pins the counting bug specifically.

---

### Verdict

Approve once item 1 is handled; it is small and self-contained. Items 2 and 3 need no code change in this PR.

I did not re-run the suite locally — my checkout is on another branch and I did not want to disturb it, so the 532-test/lint/build result is taken from your verification.

Rothermal:
Validated the re-review and fixed the remaining blocking item in e031ab0.

- soccerCanonicalStatsFromTotals now returns only projector-owned canonical stats; soc_cs is excluded at the type boundary rather than emitted as a false zero.
- Aggregate review layers soc_cs only after SOC-6B final-match clean-sheet eligibility is available.
- Parity coverage now checks both the striker and shared-clean-sheet goalkeeper across every projector-owned id, asserts aggregate goalkeeper soc_cs = 1, and asserts neither projector row has a soc_cs property.
- Pinned the migration 047/RPC game.date contract to PostgreSQL date serialized exactly as YYYY-MM-DD, not a timestamptz/full datetime.
- The dense-zero behavior for other projector-owned ids is unchanged and accepted as the non-actionable observation noted in review.

Verification remains green: 81 files / 532 tests, lint 0 errors (3 existing Fast Refresh warnings), and production build passed.

Rothermal:
## Re-review — `e031ab0`

The remaining item is resolved, and resolved the better of the two ways I suggested. No further findings from me.

**`soc_cs` is now excluded structurally rather than by convention.** `soccerCanonicalStatsFromTotals` returns `SoccerProjectorCanonicalStats = Omit<SoccerAggregateStats, 'soc_cs'>` and no longer accepts a `cleanSheet` argument at all, so the projector cannot author the id even by accident — `buildProjection`'s `Object.entries(canonical)` loop can only ever write the 32 projector-owned stats. `statsFromReviewRow` layers `soc_cs` back on after SOC-6B eligibility is known, and the spread still type-checks to the full 33-key `SoccerAggregateStats`. That is a compile-time guarantee where I only asked for a behavioral one; if someone later reintroduces `soc_cs` into the mapper, the return type rejects it.

Worth noting that `soc_cs` correctly stays in `SOCCER_CANONICAL_STAT_IDS`, so `emptySoccerAggregateStats`, `addSoccerAggregateStatsInPlace`, and `normalizeSoccerAggregateStats` keep handling all 33 — the aggregate vocabulary is unchanged and only the projector's output type narrows. That is the right split.

**The parity test now proves what it claims.** It checks the striker *and* `cloud-keeper-a` — the shared-clean-sheet goalkeeper whose divergence the previous assertion couldn't see — across every projector-owned id via `withoutCleanSheet`, asserts the aggregate keeper's `soc_cs` is 1, and asserts neither projector row carries a `soc_cs` property at all. That last assertion is the one that matters most: it pins the type-boundary decision with a runtime check, so the guarantee survives a future refactor that loses the `Omit`.

The `game.date` contract is pinned in §5 next to the RPC item shape, which is what C2 needed. Dense-zero behavior for the other projector-owned ids stays as-is, correctly — that was explicitly a no-action observation.

### Verdict

Approve. All findings across the three rounds are closed:

| Round 1 | Status |
|---|---|
| Duplicate canonical mapping (`soc_start` divergence) | Fixed — shared mapper, parity-tested on two players |
| Legacy alias map unwired | Deferred with the C3 constraint recorded in the plan |
| `ineligible_source` left quality `complete` | Fixed |
| `teamId`-null counted, non-null assertion | Fixed via type predicate |
| `M:SS` vs plan's `MM:SS` | Fixed |
| Mutating accumulator name, redundant rates, `matchIds`, merged-time semantics | Fixed / documented |
| Shootout-decided ties read as draws | Documented in the plan |

| Round 2 | Status |
|---|---|
| Projector authoring `soc_cs`; parity test blind to it | Fixed at the type boundary, with goalkeeper parity coverage |
| `isIsoDate` as a hard gate — pin the RPC date shape | Pinned in §5 |
| Sparse→dense `Player.stats` | Accepted, no action |

CI passes on `e031ab0` (52s). I did not run the suite locally — my checkout is on another branch — so the 532-test/lint/build numbers are yours plus the green CI run.

Nice slice. The engine is a clean foundation for C2.

PR #239, opened 2026-07-26
SOC-6C2: add canonical aggregate source transport

Rothermal:
## Summary
- add migration 047 with RLS-scoped keyset RPCs for team/season/tournament and player/career canonical soccer publications
- conservatively repair audited historical participant source links and preserve those links during future player merges
- retain migration 041's basketball shot-chart remount when replacing `merge_players_execute`
- add typed full-pagination transport with item-level malformed-publication isolation, per-consumer cancellation, in-flight deduplication, cooperative projection, progress, typed errors, and metrics
- add migration/transport/performance coverage and update the SOC-6C roadmap and regression docs

## Verification
- `pnpm test` (83 files, 549 tests)
- `pnpm lint` (0 errors; 3 pre-existing Fast Refresh warnings)
- `pnpm build`
- focused 50-publication / 10,000-event fixture: 596.4 ms projection and 75.7 ms largest five-publication batch in the recorded baseline

## Operator steps
1. Run `supabase/scripts/audit_soccer_participant_sources_pre_047.sql` and review repairable/unprovable counts.
2. Apply `supabase/migrations/047_soccer_canonical_aggregate_sources.sql` and review its repair notices.
3. Re-run the audit, then manually verify RLS visibility, equal-timestamp pagination, one audited historical repair, one unprovable null, and a new player merge.
4. Repeat basketball regression 4a.13 and confirm duplicate-player shot rows remount to the survivor before deletion. The preservation is forward-only; shots lost before migration 041 cannot be reconstructed.

Static migration tests validate the SQL contract but do not replace the manual Supabase checks above.


Rothermal:
## Code review — SOC-6C2 canonical aggregate source transport

Reviewed against `PLAN_SOC_6C_CANONICAL_AGGREGATES.md` §5, §10, §12 (C2 scope/acceptance) and §13, plus the existing 041-046 migrations and the C1 engine.

### What this does

Adds migration 047 (two RLS-scoped keyset RPCs, a private snapshot-completeness helper, a private participant source map, a conservative historical repair, and a replacement `merge_players_execute`), the pre-047 audit script, and a typed client transport with full pagination, cancellation, in-flight dedupe, cooperative projection batching, progress, typed errors, and metrics.

Several things are done carefully and are worth calling out because they are the parts most often gotten wrong:

- **Keyset pagination is correct.** Row-wise `(publication.finalized_at, publication.id) < (p_before_finalized_at, p_before_publication_id)` handles equal timestamps properly, and — the bit I specifically checked — `nextCursor` is built from `page_row = v_limit`, i.e. the last *returned* row, with `page_row = v_limit + 1` used only as the has-more probe. Deriving the cursor from the extra row instead would silently skip one publication per page boundary; this gets it right.
- **`security definer` + explicit `public.can_read_game(game.id)` inside the CTE** is the correct pattern: the definer bypasses RLS, and authorization is re-imposed per row rather than assumed.
- **`exists (…source_player_id = p_player_id)` rather than a join** avoids the duplicate-publication multiplication when two participants resolve to the same player, satisfying that acceptance line by construction. The supporting index already exists (`idx_game_participants_source_player`, 043), and the new partial index matches the keyset order exactly.
- **The repair is genuinely conservative:** recursive lineage with a cycle guard (`not merge.survivor_player_id = any(lineage.path)`), `depth > 0`, survivor must still exist *and* be on the participant game's team, and the audit script is read-only with a test asserting it contains no `update`/`delete`/`insert`.
- Grants are tight (`revoke … from public`, grant only to `authenticated`, helpers revoked entirely), and the test asserts `not.toContain('to anon')`.

---

### Findings

**1. A cancelled-but-not-yet-settled load can be handed to a new consumer, which then fails with a spurious `aborted`.** *(highest severity)*

`consumeSharedLoad`'s `finish()` aborts the shared controller when the last consumer leaves:

```ts
shared.consumerCount -= 1
if (shared.consumerCount === 0 && !shared.controller.signal.aborted) {
  shared.controller.abort()
}
```

The map entry is only removed in the `.finally()` on `executeAggregateLoad`, which cannot run until the in-flight request actually rejects. Between `controller.abort()` and that rejection settling, `loadSoccerCanonicalAggregates` still finds the entry:

```ts
let shared = clientLoads.get(key)
if (!shared) { …create… }
return consumeSharedLoad(shared, …)
```

so a **new** consumer joins a load that is already dying and receives `{ code: 'aborted' }` despite never having aborted anything.

This is not exotic — it is the exact shape of React StrictMode's double effect (cleanup aborts, effect immediately re-runs) and of the focus/route-reentry reloads §10 mandates ("Pages reload on mount, manual Refresh, and a transition back to visible/focused"). Because UIs normally treat `aborted` as "ignore, a newer request owns the screen", the visible symptom is a page stuck in its loading state with no error and no retry affordance.

Fix is small — treat an aborted entry as a miss on lookup:

```ts
let shared = clientLoads.get(key)
if (shared?.controller.signal.aborted) {
  clientLoads.delete(key)
  shared = undefined
}
```

The existing test ("shares identical in-flight work without one consumer cancelling another") covers *concurrent* consumers, which works correctly; it does not cover a consumer arriving after the abort. Worth adding that case.

**2. One malformed item fails the entire scope, while one malformed snapshot is isolated.**

`parseAggregatePage` does `value.items.map(parseTransportSource)`, and `parseTransportSource` throws `invalid_payload` on any field problem — so a single bad item rejects the whole load. Meanwhile a structurally-valid item with a garbage `canonicalSnapshot` flows through to C1 and becomes an isolated `malformed_publication` exclusion with `quality: 'partial'` (as the test shows). Same class of defect, two very different outcomes.

Plan §5 is explicit about which one it wants: "A malformed active publication is different from a transport failure. Continue aggregating healthy publications, but label the result prominently as partial." §10.7 says the same. Isolating and reporting satisfies both that and "never silently skip malformed canonical data" — failing the scope satisfies neither.

The realistic trigger is narrower than it looks but real: `requiredString` rejects empty and whitespace-only strings, and it is applied to *presentation* fields. `games.opponent_name` and `teams.name` are `not null` but nothing stops `''`, and `tracked_team_name` (043) is nullable with `coalesce(nullif(trim(…), ''), team.name)` as the fallback. One game with a blank opponent or an unnamed team therefore takes down an entire season's aggregates with "Aggregate opponent name is invalid."

Suggested: wrap the per-item parse in a try/catch that emits a `malformed_publication` exclusion (the C1 machinery already exists and metrics already count it), and relax display-name fields to nullable-with-fallback so only identity fields (`publicationId`, `finalizedAt`, `snapshotFingerprint`, `game.id`) are strict.

**3. Server-side *quality* filtering is invisible to the quality signal.**

`_soccer_canonical_snapshot_completed` decides eligibility by scanning for the terminal active control event, ordering by `(sequence)::bigint desc, ordinality desc, id desc`. The TypeScript authority orders by `period.order`, then `elapsedMs`, then `sequence`, then `id` — a different implementation of "latest event", maintained separately, in a different language.

Divergence in the permissive direction is harmless: the plan already requires the client to rebuild and verify, and C1 will exclude it visibly. Divergence in the *restrictive* direction is silent — the publication is never returned, so there is no exclusion row, no `malformed_publication`, no partial flag, and the user sees `quality: 'complete'` over an incomplete set. That is the one outcome §5's "never silently skip malformed canonical data" is written to prevent.

Note this is specifically about the *quality* predicate. `can_read_game` filtering must be silent (you should not learn that games you cannot read exist), and `status = 'final'` is a definitional scope bound. `_soccer_canonical_snapshot_completed` is different in kind: it is a judgement about data health, and the client is the declared authority on that judgement.

Cheapest mitigation: have the RPC return the terminal control-event type/reason as metadata and let the client classify, or at minimum document in §5 that this predicate must stay behaviourally aligned with the TS ordering and that false negatives are undetectable.

**4. A production basketball data-loss fix is bundled in undeclared and untested.**

The `shot_chart` remount is a real and valuable catch — migration 032 added `shot_chart.player_id ON DELETE CASCADE` after the merge RPCs (024/029) were written, so `merge_players_execute` has been silently deleting the duplicate player's entire court-shot history on every merge since. The comment in the SQL explains it well.

But it is not SOC-6C2 work, it affects **basketball in production**, it is not in the PR summary, it is not in the plan's C2 scope list, and — unlike the `game_participants` remount, which has an explicit ordering assertion — it has no test at all:

```ts
expect(mergeSql.indexOf('update public.game_participants'))
  .toBeLessThan(mergeSql.indexOf('delete from public.players where id = p_duplicate_id'))
```

The mirror assertion for `update public.shot_chart` is one line and protects the more valuable data. Beyond that, this deserves its own line in the PR body and its own regression step, because an operator reading the summary has no reason to manually verify a basketball merge. Worth stating explicitly that the fix is forward-only — shots already lost to previous merges are unrecoverable.

**5. Static test gaps around the parts that fail silently.**

`migration047.test.ts` is a sensible contract test and the ordering assertion in finding 4 is exactly the right idea. Two additions worth making, both covering regressions that no other test would catch:

- assert `where itemized.page_row = v_limit` for the cursor. Today only `expect(sql).toContain("'nextcursor'")` guards it; a change to `v_limit + 1` would drop one publication per page boundary while every unit test, the 50-publication fixture (single page), and CI all stay green;
- assert the `shot_chart` remount ordering (finding 4).

**6. Smaller items**

- `payloadBytes` is computed as `octet_length(convert_to(ranked.canonical_snapshot::text, 'UTF8'))` — a second full serialization of every snapshot purely for a metric, on the ~10 MB reference payload. Consider `pg_column_size` (cheaper, different semantics — note it if you switch) or accepting a client-side estimate.
- Two different `eventCount`s ship in one result: `metrics.eventCount` is the server-claimed count over all *fetched* sources, `aggregate.metrics.eventCount` is the real count over *included* matches. Both are per spec, but they will disagree whenever anything is excluded; worth a comment so instrumentation doesn't treat them as the same number.
- The in-flight key includes `activeRoster` via `JSON.stringify`, so two consumers passing the same roster in a different order miss the dedupe entirely. Sorting by `playerId` before serializing would make it order-insensitive.
- `canonicalSnapshot` is cast (`as unknown as SoccerCanonicalSnapshot`) after only an `isPlainObject` check. That is safe *only* because C1 wraps the rebuild in try/catch — worth a comment marking that dependency, since it is load-bearing and non-obvious.
- The page drain has no upper bound on pages or accumulated bytes. §5 does say "drain every page", so this is by design, but a scope larger than the reference fixture will hold every snapshot in memory simultaneously; a sanity cap that surfaces a typed error would fail better than an OOM.

---

### Test coverage

Strong for a slice whose most important behaviour is untestable in CI. Pagination/dedupe/metrics, the player RPC with optional filters, shared in-flight work, final-consumer cancellation, typed capability/access errors, page-envelope rejection with isolated malformed publications, pre-transport page-size validation, and the reviewed 50-publication/10,000-event fixture with an asserted `yieldControl` call count are all covered.

Gaps, all matching findings above: no consumer-joins-after-abort case (1), no per-item parse-failure isolation (2), no cursor-derivation assertion (5), no `shot_chart` ordering assertion (4/5).

The PR is appropriately honest that the static SQL tests do not prove RLS, cursor, repair, or transaction behaviour — §13 says the same, and the operator steps in the description are the right shape. Worth adding a basketball merge verification to step 3 given finding 4.

### Security

The definer/authorization pattern is right, grants are minimal, `search_path` is pinned on every function, and helpers are revoked from `public` so they are unreachable by `authenticated`. Two notes rather than defects: `_soccer_participant_source_map` is `security definer` with no internal authorization check — correct today because it is private and its only callers pre-filter with `can_read_game`, but it becomes a data-exposure primitive the moment anything grants it, so the comment should say so. And `canManage` derives from `current_team_role` per row, which is the right source; C3/C4 must keep gating diagnostic text on it as established in C1.

### Verdict

Good slice, and the pagination and repair logic — the two places where subtle bugs would be expensive and invisible — are correct. Finding 1 is the one I would fix before merge; it is a few lines and the failure mode is a hung page. Findings 2 and 3 are contract-alignment issues worth resolving before C3 routes real users through this path. Finding 4 is a process/visibility fix on an otherwise excellent catch.

Rothermal:
Reviewed each finding against the approved SOC-6C contract and pushed the accepted changes in `83161d4`.

### Updated

- **Aborted shared-load handoff:** valid. Aborted map entries are now treated as misses, and an older request's `finally` can no longer delete its replacement. Added a regression covering a consumer arriving after abort and a subsequent consumer sharing the replacement.
- **Malformed item isolation:** valid. A malformed publication item now produces a `malformed_publication` exclusion and partial quality without discarding healthy publications. Corrupt page envelopes/cursors remain transport failures. Blank team/opponent presentation labels receive product-facing fallbacks.
- **Terminal event ordering:** valid alignment concern. The server completed-only helper now mirrors the TypeScript order of period, elapsed time, sequence, and id. The server predicate remains because completed-only source selection is an explicit SOC-6C contract; the client rebuild is still authoritative for every returned publication. This limitation is documented.
- **Basketball shot-chart remount visibility:** migration 047 is retaining the production fix already shipped in migration 041, rather than introducing it here. Added the missing migration-047 ordering assertion, README/PR disclosure, and regression step 11t.11. The operator note explicitly records that previously lost shots are not recoverable.
- **Static regression gaps:** added exact cursor derivation checks for both RPCs and the shot-chart-before-player-delete check.
- **Roster dedupe:** normalized roster order before constructing the in-flight key and added reversed-order sharing coverage.
- **Metric/cast/security clarity:** documented the two event-count meanings, the C1 validation dependency behind the snapshot cast, and the authorization requirement of the private definer source-map helper.

### Kept By Design

- **Serialized payload bytes:** retained `octet_length(convert_to(...))` because the approved plan pins serialized UTF-8 bytes; `pg_column_size` measures a different value. Added an SQL comment to prevent an accidental semantic change.
- **No arbitrary page/byte cap:** deferred. The approved slice requires draining all pages and measuring before optimization. A fixed cap would make a valid large scope fail without an agreed product threshold; the transport already reports page, payload, and timing metrics for a measured follow-up.

Verification after the update:

- `pnpm test`: 83 files, 549 tests passed
- `pnpm build`: passed
- `pnpm lint`: 0 errors; the same 3 pre-existing Fast Refresh warnings

Rothermal:
## Re-review — `83161d4`

All three code findings are fixed, and each fix handles the half that is easy to miss. One correction to my previous review first.

### Correction: the shot-chart remount was not new here

You are right — `041_merge_preserve_shot_chart.sql` already ships the remount, including the same `ON DELETE CASCADE` explanation in its header comment. 047 is carrying it forward because `create or replace` replaces the whole function body, not introducing it. My "bundled undeclared production fix" framing was wrong, and the disclosure I asked for was not owed.

The part of that finding that did hold was the narrower one: nothing guarded the remount through a wholesale copy-forward, which is exactly the operation 047 performs. The new assertion closes that, and 11t.11 pointing at basketball regression 4a.13 is a better answer than the README line I suggested.

### Verified fixes

**1. Aborted shared-load handoff.** Both halves are there:

```ts
if (shared?.controller.signal.aborted) { clientLoads.delete(key); shared = undefined }
…
.finally(() => { if (clientLoads!.get(key) === current) clientLoads!.delete(key) … })
```

The second one matters as much as the first — without the identity check, the dying load's `finally` would evict its own replacement and the next consumer would start a third request. The new test walks the exact sequence (abort → replacement starts fresh → a third consumer shares the replacement) and pins `rpc` at 2 calls, which is the assertion that would fail under either half being wrong.

**2. Malformed item isolation.** `parseTransportRecord` wraps the per-item parse and emits a `malformed_publication` exclusion with a synthesized id when `publicationId` itself is unreadable, so a bad item can never be silently dropped *or* take down the scope. Display labels moved to `displayString` with fallbacks while identity fields stay strict — that is the right split, and it removes the blank-`opponent_name` failure mode specifically. The test now mixes a healthy item, a bad envelope, and a bad snapshot in one page and asserts `includedMatchCount: 1` with both exclusions present, which proves the two malformed classes converge on the same visible outcome.

**3. Terminal-event ordering.** I checked this against the source rather than the description: the SQL now orders by `period.order` desc, `elapsedMs` desc with missing → `9007199254740991`, `sequence` desc, then `id` desc — which mirrors `compareGameEvents` (`stream.ts:24-31`) exactly, *including* the `elapsedMs ?? Number.MAX_SAFE_INTEGER` convention at lines 26-27. That specific constant is the part I would have expected to drift; matching it means a lifecycle event with a null clock time sorts identically on both sides. Keeping the server predicate with the client rebuild authoritative is a fair reading of the §5 contract now that the limitation is written down in the plan.

**4. Static test gaps.** The cursor assertion is better than what I asked for:

```ts
expect(sql.match(/where itemized\.page_row = v_limit\s*(?:\r?\n)/g)).toHaveLength(2)
```

The trailing `\s*\n` is what makes it work — it excludes `page_row = v_limit + 1`, and `page_row <= v_limit` cannot match either, so the count of exactly 2 pins one correct derivation per RPC. A regression to the probe row fails this test and nothing else, which was the point.

**5. Roster dedupe.** Normalized before keying *and* before the load, with a reversed-order sharing test. Worth noting the shared result stays correct for both consumers because C1's roster handling is order-independent (final ordering comes from `compareSoccerAggregatePlayerRows`), so sharing across differently-ordered rosters is safe rather than merely deduplicated.

### Kept by design — both reasonable

- **`octet_length(convert_to(…))`**: agreed, the plan pins serialized UTF-8 bytes and `pg_column_size` answers a different question. The SQL comment is the right guard against a future "optimization" silently changing what the metric means.
- **No page/byte cap**: fair. An arbitrary cap without an agreed product threshold converts a valid large scope into a hard failure, and the metrics you now emit are what a measured threshold should be derived from. Worth revisiting once C3 has real scope sizes.

### Remaining observations — no action needed

- A malformed item sharing a `publicationId` with a healthy one is resolved first-wins by page order, so a corrupted duplicate arriving first would shadow the healthy row and report partial. A server returning the same publication twice is already a contract violation, and the outcome is visible either way.
- Progress can now report `publicationCount: 10` with `projectionTotal: 9` when an item is malformed. That is more truthful than the old behaviour; just something C3's progress UI should expect rather than treat as a mismatch.

### Verdict

Approve. Findings 1-3 are fixed at the right level rather than patched at the symptom, the tests target the specific regressions rather than the happy path, and my finding 4 was my error.

Still gated on the manual Supabase matrix (RLS, equal-timestamp cursors, the audited repair, an unprovable null, and now the basketball merge at 11t.11) — the static SQL tests prove the contract text, not PostgreSQL's behaviour, which §13 already says. As before, I did not run the suite locally.

PR #240, opened 2026-07-26
SOC-6C3: add canonical aggregate destinations

Rothermal:
## Summary
- route Soccer Season Leaderboard, Team Stats, and Tournament Stats through canonical migration-047 publications
- add shared best-effort active-roster loading, cancellation, progress, guarded manual/focus refresh, stale-scope suppression, and typed backend/access/error states
- add compact category ranking for every canonical soccer stat and reviewed rate, plus Team/Tournament Overview, Players, and Games tabs
- preserve zero-appearance active roster rows, historical contributors, navigation context, tournament placement, and all existing non-soccer RPC paths
- expose generic partial-quality notices to every reader while limiting affected-game diagnostics to current team managers

## Verification
- `pnpm test` (86 files, 559 tests)
- `pnpm lint` (0 errors; 3 pre-existing Fast Refresh warnings)
- `pnpm build`
- local Vite smoke check returned HTTP 200

## Manual review
1. Open Soccer Season Stats and confirm Attack / Goals is the default, readable-team scope is labeled, Team shortcuts are identified as navigation context, and category tables remain usable on mobile.
2. Open Soccer Team Stats and Tournament Stats and compare Overview, Players, and Games against finalized canonical matches.
3. Verify a current active roster player with no appearance shows at 0 APP, while a historical contributor remains present.
4. Compare partial-quality output as owner/admin and scorer/viewer; only managers should see detailed affected-game diagnostics.
5. Refresh manually, regain focus during an active load, switch scopes, and return focus after completion; identical work must not restart and stale work must not publish.
6. Fail only the active-roster read and confirm canonical totals remain visible with a zero-appearance warning.
7. Repeat Basketball Season, Team, and Tournament routes to confirm their existing RPC-backed surfaces are unchanged.

No new Supabase migration is required; this slice consumes migration 047 applied for SOC-6C2.


Rothermal:
## Code review — SOC-6C3 canonical aggregate destinations

Reviewed against `PLAN_SOC_6C_CANONICAL_AGGREGATES.md` §7, §10, §12 (C3 scope/acceptance) and the C1/C2 layers this consumes.

### What this does

Routes soccer Season Leaderboard, Team Stats, and Tournament Stats through the migration-047 canonical path: a lifecycle hook (`useSoccerAggregateDestination`), a pure presentation model (`aggregateDestinations.ts`), and a shared component with Overview/Players/Games. Non-soccer paths keep their existing RPCs.

Things done right that are easy to get wrong:

- **Manager-gated diagnostics are correct.** `soccerAggregateManagedDiagnostics` filters on `exclusion.canManage`, which C2 derives per item from `current_team_role(game.team_id) in ('owner','admin')` — so it is genuinely per *affected* team, not per viewer. `item.message` (which carries raw projector error text) renders only inside the `managed.length > 0` block, while everyone else gets the generic count. That closes the concern I raised back in C1 exactly as specified.
- **Hook placement avoids the conditional-hook trap.** The soccer branches in `TeamStats`/`TournamentStats` return a *child* component that owns the hook, rather than calling a hook behind a sport check.
- **Object-identity churn is neutralized properly.** `scope` and `teamIds` are fresh values every render, but the hook derives string `scopeKey`/`teamKey`/`loadKey` and depends only on those, so the load effect doesn't re-fire. Easy place to create an infinite loop; it doesn't.
- **Stale work is suppressed, not merely aborted.** `resultLoadKey`/`errorLoadKey` compared against the current `loadKey` means a late resolution for a previous scope can't publish even if the abort lost the race.
- All six plan-mandated states exist: no canonical matches, no resolved players, no values in category, inaccessible scope, transport failure, partial quality (plus no-team-totals and no-games).
- Tournament scope is safe from cross-team bleed because `tournaments.team_id` makes a tournament row single-team, so `game.tournament_id = p_scope_id` is already team-bounded.

---

### Findings

**1. Focus/visibility refresh has no in-flight guard.** *(highest severity)*

`useSoccerAggregateDestination`'s visibility effect debounces the focus/visibilitychange pair by 250 ms, then calls `refresh()` unconditionally. `refresh` bumps `reloadVersion`, the load effect re-runs, and its cleanup aborts the previous controller — so **regaining focus mid-load cancels the in-flight work and restarts from page 1**, resetting progress.

§10 asks for the opposite in as many words: "Guard against duplicate focus loads while one identical request is active." The asymmetry makes it look like an oversight rather than a choice — the manual Refresh button *is* guarded (`disabled={loading}`), only the automatic path isn't. On the reviewed 50-publication / ~10 MB fixture that is ~600 ms of projection plus the full page drain thrown away for an incidental tab switch.

Note the transport's in-flight dedupe can't rescue this: the hook aborts on cleanup, so `consumerCount` hits zero and the shared load dies before the replacement is created. A `loading` ref checked inside `reloadVisible` (a ref, so the listener doesn't need re-subscribing) is enough.

**2. The generic quality message counts exclusions that are excluded by design.**

```ts
const count = aggregate.exclusions.length
```

C1 deliberately does *not* let `abandoned_match` make a result partial (`exclusions.some(item => item.kind !== 'abandoned_match')`), because §2 says abandoned matches "remain reviewable in Summary but do not contribute to first-release aggregates". The message counts them anyway, so a season with five abandoned matches and one unresolved participant tells the user "6 canonical matches or player contributions could not be included" when exactly one is a problem.

It only surfaces once something else has already made the result partial, which is why the fixture (one malformed + one unresolved, both partial-inducing) doesn't catch it. Counting `exclusions.filter(item => item.kind !== 'abandoned_match')` keeps the number consistent with whatever made the result partial in the first place.

**3. Two of the three route-contract tests can pass vacuously.**

```ts
expect(source.indexOf("teamData.seasons.sport === 'soccer'")).toBeLessThan(
  source.indexOf("rpc('get_team_game_log'")
)
```

If the guard string ever changes, `indexOf` returns `-1`, which is less than any real index — so the assertion passes while proving nothing. Only the Leaderboard test asserts its `scope={{…}}` string with `toContain`; none of the three assert the *guard* exists. The `migration047` tests got this right by pairing each ordering check with a `toContain` first; worth mirroring that here, or asserting both indexes are `>= 0`.

**4. On the soccer season Leaderboard, the team selector no longer filters what is displayed.**

The destination is season-scoped (`scope={{ type: 'season', id: selectedSeasonId }}` with `teamIds` spanning every readable team), so switching the team dropdown changes only `teamIdForLinks`. The rendered table is identical. That is the correct scope per §7's readable-teams model, and "Includes completed canonical matches from teams you can read" does communicate it — but an interactive selector that visibly does nothing reads as a broken filter.

Also worth a second look: the whole canonical section is gated on `selectedTeam && … && selectedSeasonId`, so a season with no team selected renders nothing at all despite the aggregate being season-scoped and not needing a team. Either scope by team when one is selected, or decouple rendering from `selectedTeam` and mark the selector as affecting navigation only for soccer.

**5. A roster-read failure fails the entire destination.**

`loadActiveRoster` throws into the same `catch` as the aggregate load, so a `team_players` failure yields a full-page error state even though canonical aggregates are independent of it and would have rendered fine. Zero-appearance roster rows are a supporting feature (§4), not a precondition — degrading to "aggregates without zero-appearance rows" fits the plan's distinct-states approach better than a hard failure. Low likelihood in practice (RLS filters rather than errors on select), but the blast radius is the whole page.

**6. Smaller items**

- `sortSoccerAggregatePlayers` falls through to `compareSoccerAggregatePlayerRows(left, right)` with its default `sortId` of `soc_goal`, so ties in Discipline or Goalkeeping break on goals, then assists, then SOT. Deterministic but surprising; the function already accepts a `sortId` that could be passed when the selected metric is a canonical stat.
- When Supabase is unconfigured the load effect early-returns without setting an error, so the component renders a bare title row and Refresh button with no message. Unreachable on these cloud-only routes, but it is a silent blank rather than a state.
- `setUseRpc(true)` in the soccer `TeamStats` branch is dead — the branch returns before anything reads it.
- `loadActiveRoster` re-queries the roster on every refresh, including focus refreshes, and is outside the transport's dedupe map. Minor, but it compounds finding 1.

---

### Test coverage

Good for the pure model: category/column caps across every metric, goals-first ordering with the reviewed tie-breakers, zero-appearance visibility, and manager-only diagnostic exposure. The route tests are a reasonable cheap proxy for "soccer never reaches the legacy RPC", modulo finding 3.

Gaps worth closing, all matching findings: no abandoned-exclusion case for the quality message (2), nothing covering the focus-during-load path (1), and no coverage of a roster failure with healthy aggregates (5). The manual matrix in the description covers the role/scope/mobile items that genuinely need a browser, and 11u.1 pins the readable-team boundary — that split is right.

### Security and performance

No new backend surface; this consumes 047 as stated. The diagnostic gating is the security-relevant part and it is correct, including that the generic count itself leaks nothing (every counted game already passed `can_read_game`). Performance is dominated by the transport, with one caveat: finding 1 turns an incidental tab switch into a full re-drain, which is the largest avoidable cost on this path.

### Verdict

Solid slice, and the parts with real correctness or privacy consequences — diagnostic gating, stale-scope suppression, hook placement, legacy-path isolation — are all right. Finding 1 is the one I would fix before merge since the plan names it explicitly and the fix is a few lines. Findings 2 and 3 are small and worth taking now; 4 is a product call; 5 and 6 are polish.

Rothermal:
Reviewed every finding and pushed the accepted updates in `726e580`.

### Updated

- **Focus/visibility reload guard:** valid. Automatic focus refresh now reads a synchronous loading ref and skips while an identical request is active, while retaining the 250 ms focus/visibility pair debounce. Added direct policy coverage.
- **Quality count:** valid. Abandoned matches are now excluded from both the generic partial count and manager diagnostic list because they are intentionally outside first-release aggregates.
- **Route-contract assertions:** valid. All three tests now assert that each guard exists before checking its position relative to the legacy RPC.
- **Season team control:** valid UX concern. The season-wide canonical scope remains unchanged, but the control is now labeled **Team shortcuts** and explicitly states that it affects Team Stats and player-link context only. Rendering is no longer unnecessarily gated on a selected team when the soccer season scope is otherwise known.
- **Roster failure behavior:** valid. Active-roster loading is now best effort. A roster failure continues with canonical publications and shows a warning that zero-appearance players may be missing. Added a dependency-injected loader test for this path.
- **Category tie breakers:** valid. Each category now owns relevant tie-break metrics; Discipline and Goalkeeping no longer fall through to goals/assists.
- **Unconfigured state and dead assignment:** added an explicit Supabase-not-configured state and removed the unused soccer `setUseRpc(true)`.

### Kept By Design

- **Roster reads on refresh:** retained intentionally. A completed manual or focus refresh should include newly active/inactive roster members immediately. The new in-flight focus guard removes the costly duplicate restart, and a roster-query failure no longer blocks canonical totals.

Verification after the update:

- `pnpm test`: 86 files, 559 tests passed
- `pnpm lint`: 0 errors; the same 3 pre-existing Fast Refresh warnings
- `pnpm build`: passed

Rothermal:
## Re-review — `726e580`

All six findings are addressed, and the tests added would each fail under the previous behaviour — which is the part that matters. Nothing substantive left from me.

### Verified fixes

**1. Focus refresh guard.** The policy is extracted to a pure `shouldAutoRefreshSoccerAggregates({ loading, visible, now, lastRefreshAt })` and fed from a `loadingRef` that `setLoadingState` writes synchronously. Two details make this actually correct rather than nominally correct: `load()`'s first statement runs synchronously on invocation, so the ref is already `true` before the effect returns; and the superseded load's `finally` is gated on its own `current` closure, so a late-rejecting aborted request can't clear the ref out from under the replacement. Also right that `lastReloadAt` is *not* advanced when the predicate fails on `loading` — a focus event during a load is dropped rather than queued, and the next one after completion fires immediately.

**2. Quality count.** Both the count and the manager diagnostic list now exclude `abandoned_match`. Extending it to the diagnostics list is a fair reading — abandoned matches are §2-intentional and stay reviewable in Summary, so there is nothing for a manager to act on. The fixture change is the good part: the added abandoned exclusion carries `canManage: true`, so a partial fix that filtered only the count would fail the `soccerAggregateManagedDiagnostics` assertion. That is the right way to build that fixture.

**3. Route-contract assertions.** All three now `toContain` the guard before comparing positions, so the `-1 < index` vacuous pass is closed.

**4. Season team control.** "Team shortcuts" plus "This selection controls Team Stats and player links only" resolves the misleading-filter problem directly, and the section is no longer gated on `selectedTeam`. Broadening to `sport?.id === 'soccer' || scopedSport?.id === 'soccer'` covers both the sport-param and selected-team entry paths without leaving a hole — with no team selected and no param, the legacy effect still early-returns on `!selectedTeamId`, so nothing fires.

**5. Roster best-effort.** `loadSoccerAggregateDestinationData` rethrows on abort, otherwise warns and continues with `activeRoster: []`. Dependency injection for the test is the pragmatic call here, and asserting `activeRoster: []` reached the aggregate loader pins the degradation rather than just the warning string.

**6. Tie-breakers.** Per-category `rankingMetricIds`, with `attack` still `goals → assists → SOT → minutes → name` exactly as §7 specifies. The goalkeeping test is well-chosen: equal saves, and the scorer's 5 goals would have won under the old `soc_goal` fallback, so it distinguishes old from new instead of merely passing.

### Kept by design — agreed

**Roster reads on every refresh.** Reasonable now that both the cost and the risk are gone: the focus guard removes the duplicate restart, and a roster failure no longer blocks totals. Picking up newly active/inactive members on an explicit refresh is the behaviour a user would expect.

### Two cosmetic leftovers — no action needed

- `not_configured` renders the "Try again" button (only `access_denied` suppresses it), and retrying cannot help while Supabase is unconfigured.
- When the `sport` query param is absent and `selectedTeamId` came from the URL, `isSoccerDestination` is false for the render before `selectedTeam` resolves, so the legacy `get_season_stats_resolved` call can fire once for a soccer season before the effect re-runs and clears it. Wasted read-only call that self-corrects; the section itself never renders.

### Verdict

Approve. CI passes on `726e580`. Manual matrix still owns the role/scope/mobile items — particularly step 4 (manager vs scorer diagnostics) and 11u.1 (readable-team boundary), which are the two the automated tests can only approximate. As before, I did not run the suite locally.

PR #241, opened 2026-07-26
SOC-6C4: add canonical soccer player and career aggregates

Rothermal:
## Summary
- route soccer Player Profile and Career Stats through stable-player canonical publication scopes before all legacy aggregate reads
- add Participation-at-zero, all-zero category suppression, reviewed rates, canonical game links, and season/team career history
- preserve current merged-player identity and retain basketball route behavior
- complete SOC-6C roadmap, agent, and regression documentation

## Verification
- pnpm test: 87 files, 565 tests passed
- focused SOC-6C4 suite: 4 files, 22 tests passed
- pnpm build
- pnpm lint: 0 errors; 3 pre-existing Fast Refresh warnings


Rothermal:
## Code review — SOC-6C4 canonical player and career aggregates

Reviewed against `PLAN_SOC_6C_CANONICAL_AGGREGATES.md` §7, §12 (C4 scope/acceptance) and the C1-C3 layers this builds on.

### What this does

Routes soccer Player Profile and Career Stats through the stable-player canonical scopes, adds a pure `aggregatePlayerDestinations` model (identity selection, all-zero category suppression, per-player game filtering, career segmentation), extends `SoccerAggregateGame` with `seasonId`/`tournamentId`/`playerStats`, and reuses the C3 state primitives via export rename.

Things done right:

- **Scope narrowing cannot broaden visibility.** The profile passes `teamId` + `seasonId` into the player RPC, whose `(p_team_id is null or …)` / `(p_season_id is null or …)` predicates are AND-narrowing on top of `can_read_game`. Career passes neither and stays bounded by the same read check. That satisfies the acceptance line directly rather than by convention.
- **The season-name lookup avoids the identity trap.** `useSoccerAggregateSeasonNames` derives a sorted string `key` and depends on `[key]`, so passing a freshly-built array every render doesn't refetch. That is the same defensive pattern as the C3 hook, and it is the one place here where an array dep would have produced a refetch loop.
- **All-zero suppression reuses `soccerAggregateCategoryHasValues`**, so the Player/Career rule cannot drift from what C3's category tables already do, and Participation is always visible because that helper special-cases it.
- **`selectSoccerAggregatePlayer` synthesizes a zero row** when the player has no canonical publications, so the profile still renders Participation at zero — §7's requirement, handled without a special empty branch.
- Career segments keyed by `seasonId::teamId` correctly separate a player on two teams in one season and the same team across seasons.
- Route-contract tests use the corrected `toContain`-then-`indexOf` pattern from the last round.

---

### Findings

**1. The career page can no longer switch between a player's soccer and non-soccer history.** *(highest severity)*

`isSoccerDestination` is `sportParam === 'soccer'`, and the soccer branch does `setRows([])` and returns before `get_career_stats_resolved`. `sportsInData` is therefore `['soccer']`, and the sport selector only renders when `sportsInData.length > 1` — so **with `sport=soccer` there is no selector at all**.

The reverse direction is blocked too: arriving with `sport=basketball` loads only legacy rows, and soccer produces no legacy `game_stats` (aggregate sync is disabled once sport-owned setup exists), so `sportsInData` can never contain both. The result is two mutually exclusive career pages selected entirely by an inbound URL parameter, with no in-page path between them, on the one page whose purpose is the cross-game/cross-sport view.

In-app links from `PlayerProfile` always carry `sport=`, so this mostly bites bookmarks and any older link — but the failure is silent (a player with a real basketball career simply doesn't see it) rather than an error. Since `get_career_stats_resolved` is a single cheap read, the cleanest fix is to keep loading legacy rows unconditionally and branch on **`selectedSport`** rather than `sportParam`, which restores the selector and lets it contain both.

**2. `playerStats` is attached to every game row in every scope, and only player/career ever read it.**

`matchGameRow` now builds `Object.fromEntries(match.players.map(p => [p.playerId, structuredClone(p.stats)]))` for **all** scopes. Team, season, and tournament results carry a full 33-key stats object per player per game that nothing in those destinations touches — and this is newly retained data, since the per-match player breakdown was previously discarded after aggregation. A 100-game season with 18 tracked players retains ~1,800 of those objects.

The consumer only ever reads one player: `game.playerStats[identity.playerId]`. `aggregateSoccerMatches` already receives the scope, and for `player`/`career` the transport sets `scope.id` to the player id — so filtering `playerStats` to `scope.id` for those two scope types (and omitting it otherwise) is a small change that removes the overhead everywhere it isn't used.

**3. Career segment date labels depend on an unasserted ordering invariant two layers away.**

```ts
newestGameDate: games[0].date,
oldestGameDate: games[games.length - 1].date,
```

This is correct only because `aggregate.games` inherits C1's `compareMatches` date-DESC ordering. Nothing in this module states that dependency, and nothing asserts it — if that sort ever changes, the two labels silently swap. `Math.max`/`Math.min` over `games.map(g => g.date)` is the same cost and self-evident.

Related, smaller: `teamName: first.trackedTeamName` takes whichever game sorted first, so a team renamed mid-season labels the segment by its most recent name only. Probably desirable, worth being deliberate about.

**4. The profile variant loads a team roster it cannot use.**

`PlayerProfile` passes `teamIds={[teamId]}`, which triggers `loadActiveRoster` on every load and refresh. Its only effect is adding *other* players' zero-appearance rows to `aggregate.players` — and the player component never renders `aggregate.players` except through `selectSoccerAggregatePlayer` (which already synthesizes the zero row from `identity`) and `soccerPlayerAggregateGames` (which resolves to `[]` either way). Career already passes `teamIds={[]}`; the profile can do the same and drop a query per load.

**5. Minor**

- Both pages construct `identity={{ … }}` inline, so `player` and `segments` recompute on every render — `soccerPlayerCareerSegments` re-groups and re-sums the whole career each time. Harmless in practice (the season-name hook is string-keyed, so nothing refetches), but the memos aren't doing anything; depending on `identity.playerId`/`identity.displayName` or memoizing at the call site would make them real.
- `soccerPlayerAggregateGames` requires both `matchIds.has(game.gameId)` and `playerStats[playerId] !== undefined`. Those are equivalent by construction today; the redundancy is defensible as belt-and-braces but worth a comment saying which one is authoritative.

---

### Test coverage

The pure model is well covered: current-stable-identity selection, Participation-always with all-zero suppression, per-player game filtering, and career segment rebuilding from per-game totals. Route contracts follow the corrected pattern for both new pages.

Gaps matching the findings: nothing covers the career sport-switching path (1), and there is no assertion pinning the game ordering that segment labels rely on (3) — a two-game fixture with dates out of order would catch a future regression there.

The C4 acceptance items that matter most — "team-filtered player, season player, and career scopes agree on shared publications" and "all five destinations agree on canonical stat ids and formulas" — are structurally satisfied (all five paths run the same C1 projection over the same catalog), but only the manual matrix actually exercises them together. That is the right split.

### Security and performance

No new backend surface; scopes narrow rather than broaden, and the manager-gated diagnostics come through unchanged from C3 via `SoccerAggregateQualityNotice`. Performance is dominated by finding 2, which is bounded but paid on every scope rather than the two that need it.

### Verdict

Good close-out for the SOC-6C program, and the parts with correctness consequences — scope narrowing, identity selection, string-keyed effects — are right. Finding 1 is the one worth fixing before merge; it silently hides real data on the page most likely to be reached from a saved link. Finding 2 is a small change with a real memory payoff. Findings 3-5 are hardening.

Rothermal:
Reviewed and updated in f306d7d.

- Finding 1 was valid. Career now derives available sports from the player's active/inactive team history, so Soccer and Basketball remain switchable in both directions. I intentionally did not load get_career_stats_resolved while Soccer is selected because SOC-6C requires the soccer route to select its canonical source before any legacy aggregate RPC. The legacy RPC runs only after selecting a non-soccer sport.
- Finding 2 was valid. Per-game playerStats is now omitted from team/season/tournament results and retains only the requested stable player's totals for player/career scopes.
- Finding 3 was valid hardening. Career segments sort their own games, compute explicit oldest/newest dates, and deliberately use the newest canonical game for the team display name.
- Finding 4 was valid. Player Profile no longer performs the unused active-roster read; the identity row already supplies its zero-appearance fallback.
- Finding 5 was applied. Player identity inputs are stabilized before memoized selection/grouping, and the matchIds/playerStats dual guard now documents identity versus retained-detail responsibility.

Added focused sport-option, scope-retention, and out-of-order date coverage plus regression matrix case 11v.9.

Verification: full suite 88 files / 567 tests passed; focused suite 24 tests passed; production build passed; lint has 0 errors and the same 3 pre-existing Fast Refresh warnings.

Rothermal:
## Re-review — `f306d7d`

All five findings are fixed, and two of the tests deserve specific credit for testing the regression rather than the happy path. Nothing new from me.

### Verified fixes

**1. Career sport switching — your approach is better than what I suggested.** Deriving `availableSports` from the player's `team_players` history (no `is_active` filter, so departed teams still count) and unioning through `careerSportOptions` restores switching in both directions *without* loading `get_career_stats_resolved` in soccer mode. That keeps the C4 acceptance line — canonical source selected before any legacy aggregate read — which my "just load the legacy rows anyway" suggestion would have quietly broken. Both lookups degrade silently on error to the previous legacy-plus-param behaviour, consistent with the roster best-effort precedent from C3, and RLS on `team_players`/`teams` means an unreadable team simply contributes no sport option rather than leaking one.

**2. `playerStats` is now scope-conditional and single-player.** Optional on the type, populated only when `scope.type` is `player`/`career`, and only for `scope.id`. Every consumer moved to optional chaining, and `compactGameLine` now takes `SoccerAggregateStats | undefined` with a graceful fallback (dead in practice, since `soccerPlayerAggregateGames` already filters on presence — fine as defensive code).

The test is the good part: it asserts a team-scope result `not.toHaveProperty('playerStats')` *and* that a career-scope result has exactly `['cloud-striker']`. Both halves pinned, so neither over-retention nor under-population can regress silently.

**3. Segment dates no longer depend on upstream ordering.** Segments sort their own games, derive newest/oldest by explicit reduce, and document the newest-game team-name choice. The fixture was deliberately rewritten out of date order (game-3, game-2, game-1) — under the old positional `games[0]`/`games[length-1]` that test would fail, which is exactly what a regression test for this should do.

**4/5.** Profile roster read dropped (`teamIds={[]}`), and `stableIdentity` memoizes on primitives plus a derived `identityTeamKey` string — the same string-key discipline used for `scopeKey`/`teamKey` elsewhere in this stack.

### Verdict

Approve. CI passes on `f306d7d`.

That closes SOC-6C end to end: C1's engine and single canonical vocabulary, C2's RLS-scoped keyset transport and audited participant repair, C3's team/season/tournament destinations, and now player/career. Across the four slices the recurring fixes were all the same shape — one authority per fact, one string key per effect, one visible state per failure — and it shows in how little the later slices had to re-litigate.

Remaining before release is the SOC-6C manual matrix, which is where the RLS, equal-timestamp cursor, historical repair, merge, and role-visibility behaviours actually get proven; the static and unit suites deliberately don't cover those. As with the previous rounds, I did not run the suite locally.

PR #242, opened 2026-07-26
docs: plan SOC-6D soccer settings and defaults

Rothermal:
## Summary
- document the approved 36-decision SOC-6D product and architecture contract
- split delivery into SOC-6D1 through SOC-6D4
- align the parent soccer roadmaps with team-scoped defaults, revision-aware sync, and setup-time snapshots

## Validation
- git diff --check
- documentation-only change


Rothermal:
## Plan review — SOC-6D soccer settings and default hierarchy

Documentation-only. Reviewed the contract against `src/lib/soccer/rules.ts`, `types.ts`, and migrations 018/030/040, since a plan's main failure mode is asserting something about the code that isn't true.

### What this does

Defines a four-layer soccer rule hierarchy (built-in → personal → team → match → immutable snapshot), generic versioned cloud storage for personal and team settings, revision-aware compare-and-swap sync with offline/conflict handling, and a four-slice delivery split.

The architecture section holds up on inspection, which is worth saying explicitly:

- `SoccerRuleLayers` really is `appDefaults / personalDefaults / seasonRules / gameOverrides` (`rules.ts:54-59`), so this is a rename plus semantics change on an existing 4-layer resolver rather than a new parallel system. The §2 call-out that `seasonRules` must become `teamDefaults` is the right instinct.
- The team-scoped decision is correct for the schema as it stands: `teams.season_id` (migration 018) makes seasons one-to-many over teams, so season-scoped soccer defaults genuinely would let one season control every team in it.
- Snapshot immutability matches the existing `SoccerMatchSetup.rulesSnapshot` authority, and it is consistent with the parallel decision in BKE-0 §4.1.
- Slice boundaries are sensible, and 6D1's exit condition — the migration applies without changing existing games — is the right gate to put first.

One dependency the plan can drop: `access_audit_events.event_type` is `text` with a regex check (`^[a-z][a-z0-9_]{2,79}$`), not an enum, and `record_access_audit_event` is `security definer` with a fixed `search_path` (migration 040). A new `soccer_settings_changed` type therefore needs **no** schema change and can be called internally from the new settings RPC. Worth stating in §6.2 so 6D3 isn't planned around a migration it doesn't need.

---

### Findings

**1. §3.1 claims to expose "every currently implemented pregame rule" but omits two fields — and adding them would break validation.** *(highest severity)*

`SoccerMatchRules` (`types.ts:65-82`) has 16 fields. §3.1 enumerates 14. The two missing ones are `extraTimeAvailable` and `shootoutAvailable`, and they are not simply forgotten — they are a **derived legacy mirror of `tieResolution`**:

- `validateSoccerMatchRules` (`rules.ts:165-170`) *rejects* any rules object where they disagree with `legacyAvailabilityForTie(tieResolution)`;
- `resolveSoccerMatchRules` calls `synchronizeTieResolution(resolved, tieResolutionWasSet || !legacyAvailabilityWasSet)` (`rules.ts:77-81`), so whichever of the two representations a layer set determines which one wins.

Two concrete failure modes follow from leaving this unstated:

- An implementer reading "every implemented pregame rule" adds toggles for the two booleans. Any saved profile where they disagree with `tieResolution` fails `validateSoccerMatchRules`, and under §4.1 that rejects **the whole layer** and silently falls back to inherited values — a settings page that appears to save but doesn't.
- Worse across layers: personal stores `tieResolution: 'draw_allowed'` while a sparse team layer stores `extraTimeAvailable: true`. `tieResolutionWasSet` is true, so synchronization runs from `tieResolution` and the team's override is silently discarded rather than winning as the later layer.

Fix is a sentence plus a schema rule: state that these two are derived from `tieResolution`, are never independently exposed, and that §6.3's "known keys" validation must **reject** them in stored personal and sparse layers.

**2. Array-valued rules have no sub-field granularity, but §3.1 and §4.1 imply they do.**

`regulationSegments` and `extraTimeSegments` are `SoccerMatchSegment[]`, and the resolver merges layers with a shallow spread (`{ ...rules, ...layer }`, `rules.ts:68-71`). A sparse layer therefore replaces the entire array or nothing — there is no way to override "period 2's duration" while inheriting the rest.

That collides with two statements: §3.1 lists "regulation segment count, labels, and durations" as though they were three settings, and §4.1 promises "per-field source metadata" with clearing an override resuming inheritance. For segments, the only honest source label is whole-array, and "reset this duration" cannot exist without either a deeper merge or a UI that always writes the complete array.

Pick one and say it: segments resolve atomically as a single field (simplest, and the current resolver already does this), or the resolver gains element-wise merging (a real change to specify in 6D1, not a UI detail).

**3. The team-scope rationale applies to basketball's existing shared settings, which are season-scoped.**

§2 argues shared soccer defaults must be team-scoped because "a season can contain multiple teams". That is exactly the shape of `seasons.team_stats_config` (migration 030), which basketball uses today for shared team-stat rules — so the problem the plan is avoiding for soccer already exists for basketball.

§10 defers "migrating basketball settings into the generic account-backed sport-settings tables" and frames it as "a future compatible change". It won't be purely compatible: moving basketball from season scope to team scope changes which games inherit what. Worth one line recording that the future migration is a **scope change** requiring its own decision, so it isn't later planned as a lift-and-shift.

**4. §6.3 defines conflict semantics for stale updates but not for concurrent creates.**

"Create requires that no row exists; update requires an exact current revision" — but two devices coming online simultaneously with anonymous-bootstrapped defaults both take the create path, and the loser gets a primary-key violation rather than the conflict code. Since §5.2 specifies a **Use Cloud / Keep This Device** resolution flow, the create collision should map to that same stable conflict code with the current revision returned, or the first-sync path has an unhandled error state.

**5. Permissions matrix row contradicts §4.2.**

"Apply match overrides while tracking" reads as editing rules mid-match, while §4.2 states settings changes "never update an in-progress, ended, parked, cloud-bound, or finalized game" and that the snapshot is fixed on Continue. Match overrides are a setup-time capability. Rename the row (e.g. "Apply match overrides during setup") so the matrix and §4.2 agree.

**6. Smaller**

- §3.3 says "non-soccer teams cannot store soccer settings" and §6.2 says the backend derives sport compatibility from the team's season relationship. Given `teams.season_id` is nullable, the plan should state what happens for a team with no season — reject, or treat as incompatible.
- §5.1's "Signing out returns immediately to anonymous defaults" and §5.2's pending-sync marker interact: a signed-in user with unsynced pending edits who signs out should have a defined outcome for those edits (retained under the account cache and retried on next sign-in, presumably). Worth one line, since §9 tests sign-out restoration but not pending-edit survival across it.

---

### Test matrix

§9 is thorough and correctly separates automated from manual. Two additions matching the findings above: a case asserting that stored layers containing the legacy availability keys are rejected (1), and one pinning segment-array override/reset semantics once decided (2). The instruction to re-run basketball settings and setup regression because shared infrastructure changes is the right call and is easy to skip otherwise.

### Verdict

Sound plan, and unusually well grounded — the existing-architecture section is accurate where I checked it, which is not the norm for planning documents. Finding 1 is the one to fix before 6D1 starts, because it is a trap an implementer walks into by following §3.1 literally, and the failure is a silent whole-layer rejection. Findings 2 and 3 are contract clarity that gets more expensive to settle later. The rest are wording.

Rothermal:
Reviewed and updated in 0a851bd.

Accepted and addressed:
- legacy extra-time/shootout availability booleans are now explicitly derived from tieResolution, never exposed, and rejected in persisted configurable layers
- regulation and extra-time segment arrays now have explicit atomic override/source/reset semantics
- the deferred basketball migration is called out as a season-to-team scope decision, not a lift-and-shift
- concurrent create collisions return the normal revision conflict result
- the permissions row now limits match overrides to setup
- pending account edits survive sign-out in the user-keyed cache and resume only for that account
- tests now cover derived-key rejection, atomic arrays, and create conflicts

Also incorporated the audit note: migration 040 already supports a regex-valid soccer_settings_changed text event through record_access_audit_event, so no audit event-type schema change is needed.

One nuance on the seasonless-team finding: migration 018 ultimately sets teams.season_id NOT NULL, so a seasonless team is not valid in the supported schema. The plan now states that the settings RPC resolves sport through that required relationship and fails closed if it is nevertheless missing or malformed.

Rothermal:
## Re-review — `0a851bd`

All findings resolved. One correction on my side first.

**You are right about `teams.season_id`.** Migration 018 line 280 is `ALTER TABLE public.teams ALTER COLUMN season_id SET NOT NULL`, after the backfill and the fallback-season step for orphans. I read only the `ADD COLUMN` at line 32 and concluded it stayed nullable — my finding 6's premise was wrong. Recording the NOT NULL fact in §6.4 *and* keeping the fail-closed behavior is the better outcome anyway: the RPC shouldn't depend on a schema invariant it can cheaply verify.

### Verified fixes

**1. Derived legacy keys.** §3.1 now states the derivation, and §6.3 makes rejection an explicit schema rule with a matching test in §9. The part I'd have flagged next if it were missing: you also changed "complete profile" to "complete **across the configurable fields**" in §1, §4.1, and the approved-decisions list. Without that, "complete" would have implied all 16 `SoccerMatchRules` fields and contradicted the rejection rule one section later. Closing the second-order inconsistency unprompted is the right instinct.

**2. Segment arrays.** "Each resolve atomically as one array-valued field", with editing allowed at the sub-field level but saving writing the complete array, and source/reset applying to the whole array. That is the option I'd have picked — it matches what `resolveSoccerMatchRules`' shallow spread already does, so 6D1 needs no resolver change for it.

Worth being aware of the user-visible consequence, though the plan does now imply it: adjusting one period's duration at the team layer pins the *entire* segment array as a team override, so later personal changes to unrelated segment fields stop inheriting through. §3.1's "reset applies to the whole segment array" covers the escape hatch, so this needs no plan change — just don't let the UI make the write look narrower than it is.

**3-5.** Basketball scope framed as a season-to-team product change needing its own Q&A; concurrent-create collisions translated into the standard conflict response with the winning revision rather than a leaked PK error; permissions row scoped to setup. All with matching test-matrix entries.

**Pending edits across sign-out.** §5.1's addition — retained in the account-keyed cache, never applied to the anonymous scope, resuming only for that account — is the specific behavior the sign-out test was otherwise going to have to guess at.

**Audit note** folded into §6.2 with the concrete reasoning, so 6D3 won't be planned around a migration it doesn't need.

### Verdict

Approve. The plan is internally consistent, its architecture claims match the code where I checked them, and the four slices each have an exit condition that can actually fail. Ready for SOC-6D1.

For 6D1 specifically, the two things I'd expect to see land first are the derived-key rejection in the persisted schema and the segment-array atomicity tests — they're the two places where the contract and the existing resolver could silently drift apart.

PR #243, opened 2026-07-26
feat: add SOC-6D1 soccer settings foundation

Rothermal:
## Summary
- add strict schema-version-1 soccer personal, team, display, and override contracts
- refactor rule resolution to built-in -> personal -> team -> match with per-field sources and whole-layer diagnostics
- add anonymous/user-keyed local cache records plus cloud result and capability parsing
- add migration 048 with generic sport-settings tables, read-only RLS, strict soccer validation, compare-and-swap write RPCs, manager-only team writes, conflict normalization, and audit events
- update SOC-6 documentation and operational guidance

## Scope
SOC-6D1 establishes the foundation only. SettingsContext sync/UI lands in SOC-6D2, and team editor plus Match Setup inheritance land in SOC-6D3.

## Operator action
After merge, apply supabase/migrations/048_soccer_settings_foundation.sql manually.

## Validation
- pnpm test: 92 files, 583 tests passed
- focused SOC-6D1 tests: 4 files, 16 tests passed
- pnpm build
- pnpm lint: no errors; 3 existing Fast Refresh warnings


Rothermal:
## Code review — SOC-6D1 soccer settings foundation

Reviewed against `PLAN_SOC_6D_SOCCER_SETTINGS.md` (as revised in #242) and the existing rules/authorization code.

### What this does

Adds the schema-version-1 settings contracts, refactors rule resolution to built-in → personal → team → match with per-field sources and whole-layer diagnostics, adds anonymous/user-keyed local caches plus cloud result and capability parsing, and lands migration 048 with generic sport-settings tables, read-only RLS, strict soccer validation, and revision-aware write RPCs.

The two risks I flagged as the things to land early in 6D1 are both handled, and handled at the strongest available level:

- **Derived-key exclusion is enforced three times.** `SoccerConfigurableRules = Omit<SoccerMatchRules, 'extraTimeAvailable' | 'shootoutAvailable'>` makes it a *type* error to put them in a layer, `parseSoccerConfigurableRules` rejects them at runtime with a specific message, and `_validate_soccer_rule_settings` rejects them server-side via `?|`. Type-boundary enforcement is the same technique that closed the `soc_cs` issue in #238 and it's the right call here too.
- **Segment atomicity is real, and the validation is layered better than the plan described.** Shape is validated per-layer in isolation (`invalid_settings`), then effective validity is validated in composition (`invalid_effective_rules`). That second stage matters more than it looks: a cross-layer `order` collision — personal regulation order 2, team extra-time order 2 — can't be caught by either layer alone, but `validateSoccerMatchRules`' combined id/order uniqueness (`rules.ts:112-118`) fires during composition, so it surfaces as a diagnostic instead of a snapshot with duplicate period orders. Two distinct diagnostic codes for the two failure modes is the right shape.

Other things I verified rather than assumed:

- **Removing `appDefaults` from `SoccerRuleLayers` is safe.** No production caller used it — `setupRules.ts:11` layers competition profiles through `gameOverrides`, and the only other call sites pass no layers. The `seasonRules` → `teamDefaults` rename has no stale callers.
- **Segment schema matches reality exactly.** `hasExactKeys(segment, ['id','label','kind','order','durationMs'])` matches `SoccerMatchSegment` field-for-field, and the group kinds match `SoccerSegmentKind = 'regulation' | 'extra_time'`. A mismatch here would have made every save impossible, and it's the kind of thing that's easy to get subtly wrong.
- **`maxAssistsPerGoal` bounds agree** across TS (`0..2`) and SQL (`< 0 or > 2`).
- **Authorization is correct end to end.** Both RLS policies gate on `has_active_app_access()`; the team policy uses `current_team_role`, which requires `accepted_at is not null` for members (`035_team_access_hardening.sql:8-31`), so pending invitees cannot read team settings — matching the plan's "accepted team membership". `revoke all` + `grant select` means write privileges are blocked at the table level *and* absent from RLS. Helpers are revoked from `public`; only the two save RPCs are granted to `authenticated`.
- **The RPCs deliver the specific commitments from the plan revision:** `select … for update` before the CAS, create-collision normalized into the standard conflict response with the winner's revision rather than a leaked PK error, sport resolved through the required `teams → seasons` join with a fail-closed raise, and audit emitted through the existing `record_access_audit_event` with changed field *names* only — computed as a symmetric key union with `is distinct from`, so it catches additions, removals, and modifications rather than just overlapping keys.

---

### Findings

**1. `_sport_settings_is_integer` accepts values the client will always reject, so a direct RPC call can persist an unloadable layer.**

```sql
select case
  when jsonb_typeof(p_value) is distinct from 'number' then false
  else (p_value #>> '{}')::numeric = trunc((p_value #>> '{}')::numeric)
end;
```

`jsonb` stores numbers as `numeric`, which handles enormous exponents — so `{"maxOnFieldPlayers": 1e999}` is type `number`, equals its own `trunc()`, and passes the `> 0` bound. The server accepts and stores it.

The TS validators use `Number.isInteger`, and `1e999` becomes `Infinity` in JS, so `isPositiveInteger` fails. On the next read `parseSoccerConfigurableRules` rejects the layer wholesale and the hierarchy falls back to inherited values with a repair diagnostic. Net effect: a caller hitting `save_user_sport_settings_revisioned` directly (it is granted to `authenticated`) can write a row that permanently fails to load until reset. Same applies to `durationMs`.

Impact is self-inflicted and fails closed, so this is hardening rather than a vulnerability — but §6.3 promises the server validates "value ranges", and this is the one place the implementation is looser than the contract it advertises. Bounding the helper (reject `abs(value) > 2147483647`, or add explicit per-field maxima) closes it in a couple of lines and keeps the two validators symmetric.

**2. `update … returning * into v_saved` has no not-found guard.**

In both RPCs the row is locked with `for update` and a revision mismatch returns early, so the update cannot miss today. But the guarantee lives entirely in that lock two dozen lines earlier; if a later refactor drops or moves it, the function proceeds to emit an audit event and return `status: 'applied'` with a null record. `if not found then` → raise or return conflict makes the invariant local and explicit.

**3. Minor observations, no action needed**

- Resetting a team override stores `{rules: {}}` rather than deleting the row, so "never overridden" and "explicitly overridden with nothing" are indistinguishable in `team_sport_settings`. Harmless for resolution — both inherit — but if 6D3's editor wants to distinguish "inheriting" from "override cleared", it will need another signal.
- The `_validate_*` helpers are `immutable`, which is correct today because they are pure in their arguments. Worth remembering if any future rule needs to consult a table, at which point they must become `stable`.

---

### Test coverage

Unusually well aligned with the contract: the tests map onto the plan's commitments almost one-to-one, including the two I called out for this slice (`rejects derived legacy availability keys in every configurable layer`, `treats each segment array as one atomic override and source`) and the three that came out of the plan review (`normalizes create collisions`, `enforces active manager team writes and emits the existing audit event`, `retains pending account edits in the account-scoped record`). Whole-layer rejection with later layers still applying is covered, as is cache isolation and fail-closed parsing of corrupt records.

The gap matching finding 1: nothing asserts the server's numeric bounds, because there are none to assert yet. Worth a `migration048` text assertion once the bound lands, since the static suite is the only automated check on that SQL.

As the PR says, migration 048 is applied manually and the static test proves contract text rather than PostgreSQL behaviour — the RLS, CAS, and audit paths still need the manual pass.

### Verdict

Strong foundation slice. The layering is faithful to the plan, the type-level enforcement exceeds it, and the authorization posture is correct at every level I checked. Finding 1 is the only thing I'd fix before merge, and it is a two-line bound; finding 2 is cheap insurance on an invariant that currently holds by construction.

Rothermal:
Reviewed and updated in e605407.

Accepted and fixed:
- Added a shared persisted-integer ceiling of 2,147,483,647 in both the TypeScript settings parser and migration 048's SQL integer helper. This rejects huge finite JSON numerics before they can become non-finite JavaScript values, including segment duration/order fields and scalar rule integers.
- Added local not-found handling immediately after both CAS updates. An unexpected miss now reloads the current row when available and returns the standard conflict result, or conflict with a null record if the row disappeared. It cannot continue into audit or return applied.
- Added focused tests for the numeric boundary and assertions pinning all four update/create not-found guards.

No change for the two minor observations:
- { rules: {} } intentionally represents a fully inherited team layer. SOC-6D3 does not currently require product-visible provenance between never-saved and explicitly-cleared empty overrides; that can be added with a separate signal if the editor later needs it.
- The validation helpers remain immutable because they are argument-pure. A future table-dependent rule must change volatility as part of that feature.

Validation: focused tests pass (12), production build passes, and lint has no errors beyond the three existing Fast Refresh warnings.

Rothermal:
## Re-review — `e605407`

Both findings fixed, and I checked the two validators now agree field-for-field. Nothing new.

### Verified fixes

**1. Integer ceiling.** `abs((p_value #>> '{}')::numeric) <= 2147483647` in `_sport_settings_is_integer` covers every field routed through that helper — segment `order` and `durationMs`, plus `maxOnFieldPlayers`, `substitutionLimit`, `substitutionWindowLimit`, `maxAssistsPerGoal`, and `shootoutInitialKicksPerSide`. The TypeScript side covers exactly the same set: the five scalars in `validateStoredIntegerBounds`, the two segment fields in `validateStoredSegments`. Null handling matches too — `typeof candidate === 'number'` skips null, and the SQL guards with `jsonb_typeof(...) is distinct from 'null'`, so the nullable substitution limits behave identically on both sides. `abs()` on a large `numeric` doesn't overflow, so the check itself is safe against the input it's screening.

Worth noting the mechanism, because it justifies why *both* sides needed the bound rather than just the server: `JSON.parse` turns `1e999` into `Infinity` before any validator runs, so `Number.isInteger` was already rejecting that case client-side. What the TS ceiling actually closes is the opposite asymmetry — large-but-finite values like `3e9`, which `Number.isInteger` accepts and the new SQL bound now rejects. Without it, a user could enter three billion, pass client validation, and get a server rejection. Bounding both keeps the round trip honest in both directions.

int32 max is also the right ceiling for this domain: 2,147,483,647 ms is ~24.8 days, so nothing legitimate comes close.

**2. Not-found guards.** Both CAS updates now reload and return the standard conflict result, or conflict with a null record if the row vanished, and neither can reach the audit write or return `applied`. The user-scope reload correctly passes `null` for `updated_by`, matching that table's shape.

The test is the right kind: `expect(sql.match(/if not found then/g)).toHaveLength(4)` pins the exact count across both create-collision and both update paths, so deleting any one of the four fails — the same counting technique that guards the cursor derivation in 047.

One structural note, not a defect: the ceiling exists as a TS constant and an SQL literal, so "shared" means duplicated with a static text assertion as the anti-drift guard. That's the established pattern in this repo for client/server validation mirrors, and there's no way to share a literal across the two languages, so the `migration048` assertion is the right mitigation.

### Agreed on both deferrals

`{ rules: {} }` as a fully-inherited team layer is fine — provenance between never-saved and explicitly-cleared is only interesting if 6D3's editor decides to surface it, and adding a signal later is cheap. And keeping the validation helpers `immutable` is correct while they stay argument-pure; changing volatility belongs to whatever feature first needs a table lookup.

### Verdict

Approve. CI passes on `e605407`.

Remaining before 6D2 is the manual pass on migration 048 — RLS visibility for accepted versus pending membership, the CAS conflict paths under two concurrent sessions, and the audit row shape. The static suite proves the SQL text, not PostgreSQL's behaviour, which the PR already says.

PR #244, opened 2026-07-26
SOC-6D2: add personal soccer settings and reconciliation

Rothermal:
## Summary
- add the compact Settings -> Sports -> Soccer editor for all personal rule and display defaults
- add cache-first anonymous/account settings reconciliation with offline pending writes and explicit cloud conflict choices
- wire revision-aware user settings reads/writes through migration 048 and update SOC-6D documentation

## Scope
- personal defaults only; team defaults and Match Setup inheritance remain SOC-6D3
- no new migration is required beyond already-applied migration 048

## Verification
- pnpm test (93 files, 590 tests)
- pnpm build
- offline rendered smoke check at /#/settings/sports/soccer


Rothermal:
## Code review — SOC-6D2 personal soccer settings and reconciliation

Reviewed against `PLAN_SOC_6D_SOCCER_SETTINGS.md` §3, §5, §8 (6D2 scope/exit), and the 6D1 foundation.

### What this does

Adds the Settings → Sports → Soccer editor, a `useSoccerPersonalSettings` controller with cache-first reconciliation, offline pending writes and explicit conflict resolution, and wires user settings reads/writes through migration 048.

The reconciliation core is right, and it's the part I'd have expected to get subtly wrong:

- `reconcileSoccerPersonalSettings` matches the plan's precedence exactly — cloud wins on first sign-in absent a local pending edit, pending-with-matching-base becomes a CAS upload, pending-with-diverged-base becomes a conflict that preserves **both** sides, and unparseable cloud data fails closed to local with a repair error rather than overwriting anything.
- **Pending account edits genuinely survive sign-out.** On sign-out `targetScope` flips to anonymous, so `refresh` writes only the anonymous key and never reads, clears, or applies the user-scoped record; re-signing in reloads that cache and resumes reconciliation. That was the §5.1 gap I raised during the plan review and it's implemented correctly.
- The draft-sync effect (`SoccerSettings.tsx:180-189`) adopts incoming cloud values **only when the draft is clean**, so a background pull can never clobber unsaved edits — and because a dirty draft keeps its stale `draftBaseRevision`, the resulting Save produces a genuine conflict whose "cloud" side really is cloud data. I checked that specifically because a synthesized conflict side is an easy place to show the user mislabeled values; it holds up.
- `requestRef` correctly discards superseded continuations, and Save/Discard, dirty state, section and full reset, the effective preview via `resolveSoccerSettingsHierarchy`, and `display.fieldFlipped` kept separate from `rules` all match §3.

---

### Findings

**1. The controller is mounted in `SettingsProvider`, so soccer settings cloud traffic now runs app-wide for every signed-in user.** *(highest severity)*

`SettingsProvider` wraps the entire app (`App.tsx:147`), and it now calls `useSoccerPersonalSettings()` unconditionally. Three consequences, none of them scoped to the soccer settings page:

- Every signed-in user issues a `user_sport_settings` select on app boot — including users who only track basketball. `DEFAULT_SETTINGS.enabledSports.soccer` is `false` (`settingsStorage.ts:16`), and the plan keeps soccer development-only until SOC-6E, so this is unconditional cloud traffic on behalf of a sport that isn't exposed.
- Because a missing row resolves to `upload_local` (see finding 3), those users also get a **write** creating a `user_sport_settings` row containing pure built-in defaults.
- The `focus` and `online` listeners (`useSoccerPersonalSettings.ts:1180-1189`) are app-global, so every window focus anywhere — including mid-match on the tracker — triggers a soccer settings reconcile that can itself perform a write.

§5.1's "loads cached settings immediately and reconciles in the background" justifies background reconciliation, but not doing it for a disabled sport on every screen. `isSportEnabled('soccer')` is already available in that very context and is the natural gate; alternatively give the hook an `enabled` flag and let the settings page turn it on, keeping the context shape unchanged.

**2. The auto-refresh path has no in-flight guard, which can manufacture a conflict against the user's own edit.**

`refresh` is called directly by the focus/online listeners with no debounce and no in-flight check. `requestRef` prevents *stale state commits* but not overlapping requests, and one of those overlaps has a bad outcome:

1. Refresh A reads cloud revision 5, sees pending `baseRevision: 5`, and starts its upload.
2. A focus event starts refresh B (`requestRef` → 2). B reads the cloud before A's write lands, so it also sees revision 5 with pending base 5 and decides to upload.
3. A's write lands (revision → 6); A's continuation bails because it's superseded, so nothing records the new revision.
4. B writes with `expectedRevision: 5` → server returns `conflict`.
5. The user gets a conflict dialog whose two sides are **identical** — their own edit versus their own just-uploaded edit — and "Keep This Device" then bumps to revision 7 for nothing.

No data is lost, but it undermines the one dialog that has to be trustworthy. This is the same class of issue I raised on SOC-6C3, and you already built the fix there: a synchronous `loadingRef` plus the pure `shouldAutoRefreshSoccerAggregates({ loading, visible, now, lastRefreshAt })` policy. The same shape applies here, and it matters more because a focus event on this path can trigger a write rather than just a read. It also compounds with finding 1, since the listeners are currently app-wide.

**3. A first visit creates a cloud row from built-in defaults, unprompted, and reports `saving` while doing it.**

With no cloud row and no account cache, `reconcileSoccerPersonalSettings` returns `upload_local` with `bootstrapSettings` — which falls back to `DEFAULT_SOCCER_PERSONAL_SETTINGS` when there's no anonymous cache either. `refresh` then commits `status: 'saving'` and uploads.

So a user who has never customized anything gets a persisted row of pure defaults, and (per finding 1) that happens on app boot rather than on visiting the page. §5.1 says anonymous defaults *may* initialize a cloud profile when no record exists — uploading built-ins when there is nothing to preserve is a broader reading, and it also means offline first-boot leaves a pending marker for settings the user never touched. Consider uploading only when there is an anonymous profile worth migrating, or deferring creation to the first real save.

**4. `dirty` uses order-sensitive `JSON.stringify`, but jsonb returns keys sorted.**

`dirty` compares `JSON.stringify(draft)` against `JSON.stringify(soccerSettings)`. Postgres `jsonb` does not preserve key order, so cloud-sourced settings come back alphabetically ordered while `structuredClone(DEFAULT_SOCCER_PERSONAL_SETTINGS)` carries declaration order. "Reset all" (`SoccerSettings.tsx:544`) therefore marks the form dirty even when the stored settings are already semantically identical to defaults, and Save then bumps the revision for a no-op write.

Everyday editing is unaffected because `draft` is always cloned from `soccerSettings`. The repo already has the fix: `canonicalizeJsonLike` / `stableJson` in `gameEvents/stream.ts` exist for exactly this comparison problem.

**5. Minor: `createSoccerSettingsCacheRecord`'s `pendingBaseRevision` encodes three states on one parameter.**

`undefined` means "no pending", `null` means "pending create", and a number means "pending update". Every call site I checked uses it correctly, but the distinction between `undefined` and `null` is invisible at the call site and getting it wrong silently marks a synced record as pending. A discriminated `pending: { baseRevision: number | null } | null` option would make the intent explicit.

---

### Test coverage

`personalSettingsSync.test.ts` covers the reconciliation matrix well — the decision function is the right thing to have extracted and tested in isolation, and the cloud parsing additions are covered too.

The gaps line up with the findings, and all three are testable without a browser because the logic is already in pure or hook-shaped units: no case for two overlapping refreshes (2), none for the defaults-only bootstrap deciding to upload (3), and none pinning `dirty` across a key-reordered round trip (4). Finding 1 is best addressed by a test asserting no cloud call happens when soccer is disabled.

### Security and performance

No new backend surface; 6D1's RLS and CAS RPCs do the enforcement and this slice consumes them correctly with expected revisions threaded through. The performance concern is entirely finding 1 — an unconditional read plus a one-time write per user, and a reconcile per window focus app-wide, for a sport that ships disabled.

### Verdict

The reconciliation logic and the editor are solid, and the two subtle correctness points I went looking for — sign-out pending survival and the conflict dialog's cloud side — both check out. Finding 1 is what I'd fix before merge: it changes app-wide behaviour for every user on behalf of a disabled sport, and the gate is one condition in a context that already exposes it. Finding 2 is a known pattern with an in-repo fix and should ride along. Findings 3-5 are smaller.

Rothermal:
Reviewed all five findings against the current implementation and pushed fixes in 9e282a0.

- **App-wide cloud traffic:** valid and fixed. Reconciliation now activates only when Soccer is enabled or Settings -> Sports -> Soccer is open; focus/online listeners follow the same gate.
- **Overlapping refresh self-conflict:** valid and fixed with a synchronous single-flight refresh guard, plus policy coverage for disabled and already-running states.
- **Unprompted built-in-default rows:** valid and fixed. No cache + no cloud row now remains local and does not write; an actual anonymous/account cache still bootstraps a missing row.
- **Order-sensitive dirty detection:** valid and fixed with canonical JSON fingerprints; reordered-object coverage was added.
- **Ambiguous pending option:** accepted and cleaned up. Cache creation now takes explicit pending: { baseRevision } | null.

The SOC-6D plan now records the scoped/single-flight lifecycle. Verification: 94 test files / 594 tests pass, production build passes, and lint has no errors (only the three existing Fast Refresh warnings).

Rothermal:
## Re-review — `9e282a0`

Four of five findings are cleanly fixed. One is fixed for the case I described but not for the adjacent case that reaches the same outcome, and the gating fix introduced one thing worth a look.

### Verified fixes

**1. App-wide cloud traffic — fixed.** `cloudEnabled = Boolean(settings.enabledSports.soccer) || location.pathname === sportSettingsPath('soccer')`, threaded through `initialState`, the `refresh` early return, the mount effect, and the focus/online subscription. A basketball-only user on any other screen now does no soccer settings I/O at all.

**3. Unprompted default rows — fixed, and the migration path is preserved.** `reconcileSoccerPersonalSettings` now returns `use_local` when there is neither a cloud row nor a local record, and the hook commits `cache: null` so nothing is written. The `reconciliationCache` synthesis still bootstraps a genuine anonymous profile into the cloud when the account has no row, so the §5.1 behaviour survives while the empty case stays local.

**4. Order-sensitive dirty detection — fixed.** `stableJson` exported from `gameEvents/stream.ts` and wrapped as `soccerSettingsFingerprint`, used for both `dirty` and the draft-sync effect. Reusing the existing canonicalizer rather than writing a second one is the right call.

**5. Pending option — fixed.** `pending: { baseRevision } | null` is explicit at all nine call sites.

Also verified the single-flight guard itself is correctly written: the flag is set *after* the `shouldStart` check, so a rejected call can't clear another call's flag, and the `finally` covers every early return inside `refresh` including the awaited write.

---

### Findings

**1. The guard covers refresh-vs-refresh but not write-vs-refresh, so the self-conflict race is still reachable.**

`refreshingRef` is set only by `refresh`. `save` and `keepDevice` call `applyCloudWrite` directly without touching it, so a focus event during a deliberate save walks the same path as before:

1. `save(draft, 5)` → `requestRef` = 1, caches `pending { baseRevision: 5 }`, starts `applyCloudWrite(..., 5, ..., 1)`. `refreshingRef` is still `false`.
2. Focus fires → `refresh()` passes `shouldStartSoccerSettingsRefresh(true, false)`, takes `requestRef` = 2, reads the cloud before the save's write lands (still revision 5), sees pending base 5, and decides `upload_local` with `expectedRevision: 5`.
3. The save's write lands (revision → 6); its continuation bails on `requestId(1) !== requestRef.current(2)`, so the successful save is never reflected in state.
4. The refresh's write with `expectedRevision: 5` returns `conflict` — identical sides again, and "Keep This Device" then bumps to revision 7 for nothing.

Clicking Save and alt-tabbing during the round trip is an ordinary thing to do, so this is the same user-visible symptom in a slightly narrower window. Making the flag mean "a cloud write or read is in flight" rather than "a refresh is running" closes both: set it in `save`/`keepDevice` as well, or move it into `applyCloudWrite`, and give `shouldStartSoccerSettingsRefresh` a third `writing` argument so the policy test can pin it the way the current two cases are pinned.

**2. `SettingsProvider` now calls `useLocation()`, so the app-wide provider re-renders on every navigation.**

This is *safe* — `HashRouter` wraps `App` in `main.tsx:9-11`, so the hook has a Router ancestor and won't throw. But the context value is built inline, so it is a fresh object on every render, and `SettingsProvider` sits above `GameProvider` and every route. Previously it re-rendered only when settings changed; now every route change re-renders all settings consumers, to gate a development-only feature.

Practical impact is probably modest, since navigation already re-renders the route subtree. If you want it tighter: gate on `enabledSports.soccer` alone and let the settings page turn cloud reconciliation on through its own effect or a prop, or read the location in a small child that only computes the flag. Memoizing the context value would also stop the cascade at the provider.

---

### Test coverage

`shouldStartSoccerSettingsRefresh` is a good extraction and both its states are pinned. The reconciliation additions (`use_local`) and the reordered-object fingerprint case are covered too.

The gap is finding 1: the policy function can't express the write-in-flight case yet, so nothing can test it. Adding the third argument makes it a two-line test alongside the existing two.

### Verdict

Close. Finding 1 is a small change to an existing guard and closes the last path to a conflict dialog the user can't act on sensibly; I'd take it before merge. Finding 2 is a judgement call about how much re-render surface a dev-only gate should buy, and it's safe either way.

CI passes on `9e282a0`.

Rothermal:
Re-reviewed the two follow-up findings and pushed 416bef.

- **Write-versus-refresh race:** valid and fixed. pplyCloudWrite now marks every cloud write in flight synchronously and clears it in inally; refresh policy checks both read/reconcile and write activity. The policy test now pins disabled, refreshing, and writing states.
- **Provider-wide location subscription:** safe but avoidable, so updated. SettingsProvider no longer calls useLocation; SoccerSettings activates reconciliation only for its mounted lifetime through a stable context setter. Enabled Soccer still activates it globally for future D3 consumers.

Verification remains green: 94 test files / 594 tests, production build, and lint with only the three existing Fast Refresh warnings.

Rothermal:
## Re-review — `b416bef`

Both items fixed. Approve.

### Verified fixes

**Write-versus-refresh race.** `applyCloudWrite` now sets `writingRef` synchronously — before the first `await`, so the flag is up before control yields — and clears it in `finally`; `refresh` checks both flags. Re-running the sequence from my last review, it's blocked at step 2: the save marks the write in flight, the focus event's `refresh` sees `writing === true` and returns without reading, and the save's own continuation commits revision 6 normally. The next focus then sees a non-pending cache and takes `use_cloud`. No stale-expectation write, no identical-sided conflict.

The policy test now pins all three states, so `cloudEnabled`, `refreshing`, and `writing` each have a case.

**Provider-wide location subscription.** `useLocation` is gone; `SettingsProvider` holds a `soccerSettingsPageActive` boolean that the page raises and lowers from a mount effect through the raw `useState` setter. Since that setter is referentially stable the child effect runs exactly once per mount, and the provider now re-renders twice per visit to the soccer settings page instead of on every navigation. That's a better shape than what I suggested — the activation signal belongs to the surface that needs it rather than being inferred from a path string, which also means SOC-6D3's team editor can opt in the same way without another path comparison.

### One hardening note, no action needed now

`applyCloudWrite`'s new early return (`if (writingRef.current) return false`) exits without reconciling the optimistic `status: 'saving'` that every caller commits just before invoking it. That path is currently unreachable: Save is disabled while `status === 'saving'`, Refresh is disabled while checking or saving, `refresh` bails on the same flag, and both conflict actions self-guard because the first click clears `conflict` before the second can run.

It stays unreachable only as long as that remains true. If SOC-6D3 adds a programmatic `saveSoccerSettings` caller — it is exposed on the context — hitting that branch would leave the UI at `saving` until the next focus or online event triggers a recovering refresh. Cheapest insurance when that happens is to have the early return commit `pending` rather than returning silently.

### Verdict

Approve. CI passes on `b416bef`.

All five findings from the first pass and both follow-ups are closed, and the two that mattered — app-wide cloud traffic for a disabled sport, and a conflict dialog the user couldn't act on sensibly — are gone. Remaining before 6D3 is the manual matrix: two-device conflict resolution, offline pending then reconnect, sign-out with a pending account edit, and the multiple-accounts-on-one-device cache isolation case.

PR #245, opened 2026-07-26
SOC-6D3: add team defaults and setup inheritance

Rothermal:
## Summary
- add cache-first, revision-aware shared soccer team defaults with owner/admin editing and scorer/viewer read-only access
- add compatible-team copy and personal/team settings shortcuts in Team Manage
- resolve built-in, personal, team, and sparse match layers in Soccer Match Setup with per-field sources and fixed snapshots
- preserve explicit in-session match overrides across team changes and retain existing setup snapshots until deliberate edits
- update SOC-6D status, architecture guidance, and regression coverage

## Backend
- no new migration is required; migration 048 already provides team settings RLS, sport validation, CAS writes, and audit events

## Verification
- pnpm test: 95 files, 601 tests passed
- pnpm build: passed
- pnpm lint: passed with 3 pre-existing Fast Refresh warnings


Rothermal:
## Code review — SOC-6D3 team defaults and setup inheritance

Reviewed against `PLAN_SOC_6D_SOCCER_SETTINGS.md` §3.3, §4.2, §5.2, §8 (6D3 scope/exit) and the 6D1/6D2 layers.

### What this does

Adds shared team defaults with a Team Manage editor and compatible-team copy, a `useSoccerTeamSettings` hook, and four-layer resolution in Soccer Match Setup with per-field sources, sparse match overrides, and snapshot fixation. `SoccerRulesOverrideEditor` is extracted so personal, team, and match layers share one control surface.

The §4.2 semantics — the hardest part of this slice — are right, and I checked each clause:

- **Match overrides survive a team change** because `matchOverrides` is independent state and `setupRules` simply re-resolves against the new `selectedTeamSettings`.
- **An existing `rulesSnapshot` stays fixed until deliberately touched.** `resolveSoccerSetupRuleState` returns the snapshot verbatim as `rules` while presenting it as a diff for editing, and `updateMatchOverrides` releases the snapshot only on an actual edit.
- **A deliberate edit converts the snapshot losslessly.** The editor emits `soccerRulesOverrideFromDifference(inherited, next)` — a *complete* difference set — so the snapshot's other deviations become explicit overrides instead of being silently dropped on first edit. That's the failure mode I went looking for and it isn't there.
- **`handleContinue` refuses to snapshot until the selected team's defaults have loaded** (`scopeTeamId !== selectedTeamId`, or status idle/loading). Without that guard a fast Continue would fix a half-resolved hierarchy; good catch on your side.

Authorization and offline posture also check out:

- `mayEdit={mayManageRoster}` reuses the established owner/admin authority rather than inventing a parallel check, and copy options are filtered to other **soccer** teams where `teamRolesById[team.id]` is set. RLS on `team_sport_settings` is the real gate, and sport compatibility is transitively enforced because 048 only lets a soccer team store soccer settings — so the client filter is convenience, not security.
- Team cache records are always written with `pending: null`, and `save` refuses when `!navigator.onLine`, with an `offline` listener that flips status to `cached` and says why. That matches §5.2's "shared editing unavailable offline, cached values still visible and still seeding setup, labeled with sync state".
- Team settings structurally cannot carry display orientation, since `parseSoccerTeamSettings` requires exactly `{ rules }` — §3.2's "team defaults cannot enforce it" holds by type rather than by convention.
- `key={selectedTeam.id}` remounts the panel per team, so no state bleeds across a team switch.

---

### Findings

**1. `useSoccerTeamSettings.save` doesn't participate in the request guards, so a focus event during a save can roll the UI back to pre-save state.**

`refresh` correctly carries the 6D2 lesson — `loadingRef` single-flight plus a `requestRef` check on its continuation. `save` has neither: it never sets `loadingRef` and never bumps `requestRef`. So:

1. User clicks Save → `setStatus('saving')`, `saveTeamSportSettings(..., expectedRevision: 3)` in flight. `loadingRef` is `false` and `requestRef` unchanged.
2. Window focus → `refresh()` passes its guard, takes `requestRef` = N+1, and reads the cloud before the save lands (still revision 3).
3. The save's write lands → server revision 4; its continuation has no `requestRef` check, so it sets settings/revision to 4 and status `synced`.
4. The refresh's stale response resolves, `requestId === requestRef.current` is still true, and it **overwrites state *and* the cache** with the revision-3 record, status `synced`.

The editor now shows pre-save values at revision 3 while the server holds 4, so the save looks reverted; the next Save uses `expectedRevision: 3` and CAS-conflicts. This is the same defect class as 6D2's write-versus-refresh race, but the outcome is worse there because the stale read wins and poisons the cache.

The fix is the pattern you already built in `useSoccerPersonalSettings`: set a `writingRef` synchronously at the top of `save` (cleared in `finally`), have `refresh`'s guard check it, and bump `requestRef` in `save` so any in-flight refresh continuation is discarded. Note the button being disabled during `saving` doesn't help — the focus event starts the refresh, not a second click.

**2. `sameStoredValue` and `soccerRulesOverrideFingerprint` use order-sensitive `JSON.stringify`.**

This is the comparison problem fixed in 6D2 with `stableJson`, reintroduced in `settings.ts`. It only matters for the two array-of-object fields, whose elements are `{ id, label, kind, order, durationMs }`: an in-memory `inherited` carries declaration order, while a value that has round-tripped through Postgres `jsonb` comes back alphabetically sorted.

During editing both sides are in-memory and derived from the same object, so ordinary use is unaffected. The reachable path is `resolveSoccerSetupRuleState`'s `preservedSnapshot` — a `rulesSnapshot` from a cloud-hydrated soccer game whose event stream is still empty (setup redirects away once events exist, so it's a narrow but real window). There, `regulationSegments` and `extraTimeSegments` are reported as match overrides despite being identical to inherited, mislabeling their source; and because the next edit emits the whole difference set, those arrays then become *genuine* pinned overrides, so later team-default segment changes stop inheriting for that game.

`stableJson` is already exported from `gameEvents/stream.ts` — swapping it in is one line in each function.

**3. Minor**

- `releasePreservedSnapshot` clears `matchOverrides` as well as the snapshot. It's harmless today because the snapshot path never populates `matchOverrides`, but the two-state coupling is implicit; a comment saying why the clear is safe would keep a future edit from making it lossy.
- Confirming the 6D2 note is still safe: 6D3 adds no programmatic `saveSoccerSettings` caller, so `applyCloudWrite`'s silent early return stays unreachable through the UI.

---

### Test coverage

Good targeting on the semantics that matter: sparse-difference storage, override preservation across team change, snapshot fixation until deliberate reset, team-scoped cache isolation by account *and* team, and both identity filters on the team read. `setupSettings.test.ts` covering the three §4.2 clauses directly is the right place for them.

Gaps matching the findings: nothing exercises save-versus-refresh interleaving (1), and nothing pins the difference function across a key-reordered snapshot (2) — the latter is a two-line test now that the fixture just needs a segment object with permuted keys.

### Security and performance

No new backend surface; 048's RLS, sport validation, CAS, and audit do the work and this slice threads expected revisions through correctly. Both write paths are manager-gated in the UI and enforced server-side. Nothing performance-sensitive — the team hook is scoped by `active` and only mounts on Team Manage or cloud setup.

### Verdict

Strong slice; the setup-inheritance semantics are the subtle part and they're correct, including the snapshot-to-override conversion. Finding 1 is what I'd fix before merge — it's the same guard you already wrote next door, and the failure silently discards a manager's save to a shared authority. Finding 2 is a one-line consistency fix worth taking while it's cheap.

CI passes on the head commit.

Rothermal:
Reviewed and validated the reported items against the implementation and SOC-6D3 contract.

Implemented in b569cb1:
- serialized shared team settings writes with a synchronous writing guard, blocked focus/online refresh during writes, and invalidated stale read/write continuations through the request generation
- switched sparse-difference and fingerprint comparisons to stableJson so Postgres jsonb key ordering cannot create false segment overrides
- added regression coverage for the refresh guard and nested segment key reordering
- documented why releasing a preserved setup snapshot can safely clear the otherwise-empty matchOverrides state

Verification: 96 test files / 603 tests pass; production build passes; lint has only the three existing Fast Refresh warnings.

Rothermal:
## Re-review — `b569cb1`

Both findings fixed. Approve.

### Verified fixes

**1. Shared write serialization — closed in both directions, which is more than I asked for.**

`writingRef` and the `requestRef` bump are both set synchronously at the top of `save`, before the first `await`, and `finally` clears the flag. `refresh` checks `writing` through the new policy. I re-traced both interleavings:

- *Save then focus:* `refresh` sees `writing === true` and returns, so the stale read never starts. The save's continuation still matches `requestRef` and commits revision 4.
- *Refresh in flight, then Save:* `writingRef` is false so the save proceeds, but the `requestRef` bump invalidates the refresh's continuation, which returns before it can overwrite state or the cache.

That second direction is the one `writingRef` alone wouldn't have covered — the bump is what handles it. Also correct that `save` does *not* check `loadingRef`: a background read shouldn't block a manager's save when the read's result is already going to be discarded.

The policy test pins all four combinations, and `refresh`'s `finally` still releases `loadingRef` on the early-return path.

**2. Order-insensitive comparison — both call sites.** `sameStoredValue` and `soccerRulesOverrideFingerprint` now use `stableJson`. The new test permutes segment element keys and asserts both an empty difference and equal fingerprints, so it fails under the old `JSON.stringify` — which is what makes it worth having.

**3.** The `releasePreservedSnapshot` comment records why clearing the otherwise-empty `matchOverrides` is safe, so a future edit won't quietly make it lossy.

### Verdict

Approve. CI passes on `b569cb1`.

That's SOC-6D1 through 6D3 landed with the four-layer hierarchy behaving per §4.2 and both shared-write paths serialized. Remaining before 6D4 is the manual matrix — owner/admin versus scorer/viewer on the team editor, two browser sessions editing the same team scope, offline shared-edit refusal with cached values still seeding setup, and a soccer game parked before a settings change then resumed afterward.

One thing worth carrying into 6D4: the write-versus-read serialization has now been rediscovered three times (6C3 focus refresh, 6D2 personal writes, 6D3 shared writes), each time as a separate `shouldStart*` policy plus a pair of refs. If 6D4 adds another cloud-backed surface, a small shared hook or helper for "cache-first scope with single-flight read and serialized write" would stop the fourth rediscovery — the three existing implementations are close enough in shape to factor.

PR #246, opened 2026-07-27
SOC-6D4: harden soccer settings and document regression coverage

Rothermal:
## Summary
- contain corrupt or unavailable browser storage without crashing or weakening cloud authority
- harden Soccer settings and Match Setup keyboard, announcement, reset, and narrow-screen behavior
- expand schema/cache/cloud/migration tests and add the SOC-6D operator regression matrix
- mark SOC-6D complete and SOC-6E as the next release phase

## Backend
No new migration. Migration 048 remains required for account-synced personal settings and shared team defaults.

## Validation
- pnpm test: 97 files, 613 tests passed
- pnpm lint: 0 errors; 3 existing Fast Refresh warnings
- pnpm build: passed


Rothermal:
## Code review — SOC-6D4 hardening and regression coverage

Reviewed against `PLAN_SOC_6D_SOCCER_SETTINGS.md` §8 (6D4 scope/exit) and the 6D1-6D3 layers.

### What this does

Contains browser-storage failures, completes the accessibility pass on the settings and setup surfaces, expands schema/cache/cloud/migration tests, and adds the SOC-6D operator regression matrix.

Low-risk slice, and the parts that could have gone wrong don't:

- **Storage containment covers both directions.** `saveSportSettingsCache` now returns `{ ok, error }` instead of throwing, and `loadSportSettingsCache` already wrapped `getItem` + `JSON.parse` in a try returning `null`. Previously a quota or private-mode failure threw from inside `cacheAndCommit`, which sits in `refresh`'s `try`/`finally` with no `catch` — so it surfaced as an unhandled rejection. That's gone.
- **Cloud authority is genuinely unaffected**, which is the §8 clause worth checking rather than assuming. A cache write failure never changes what is sent to or read from the cloud; state still reflects the saved record, and the next `refresh` re-reads the truth. `applyCloudWrite` correctly still returns `true` after a successful RPC even when the cache write fails — the cloud write is what succeeded, and reporting otherwise would be the actual authority weakening.
- **Status/error pairing is coherent.** A cache failure sets the warning while leaving status `synced`, which is accurate: the shared write landed, only the local mirror didn't. All four team-hook call sites and the personal hook's `cacheAndCommit` surface it consistently.
- **The tab pattern is completed properly.** `role="tablist"`/`tab`/`aria-selected` already existed; this adds the roving `tabIndex` (`0` for active, `-1` otherwise), Arrow/Home/End handling that actually moves focus, `role="tabpanel"`, `aria-controls`, and `aria-labelledby` back to the active tab. That's the whole pattern rather than the common half of it — keyboard users get focus movement, not just a changed panel.
- `ConfirmDialog` gains `role="dialog"`, `aria-modal`, and `useId`-backed `aria-labelledby`; purely additive, so the existing consumers (team delete, merge flows) are unaffected.
- Test additions map onto the stated hardening goals: unsupported cloud schema with a coherent local cache, unknown keys at every persisted boundary, duplicate/incorrect segment identities, sparse-override clearing, malformed load/save responses, and both storage read and write failures.

---

### Findings

Both minor; nothing blocking.

**1. The Match Setup banner now surfaces the device-storage warning under team-defaults framing.**

The condition became `Boolean(teamSettings.error) || status === 'cached' | 'backend_update_required' | 'error'`, and the body renders `teamSettings.error ?? …`. Since `useSoccerTeamSettings` now sets `error` to the storage message on a cache-write failure even when the cloud load succeeded, Match Setup can show an amber "shared team defaults" banner reading *"Device storage is unavailable. Changes remain available for this session only."*

Two things are off there: the framing implies team defaults are unavailable when they loaded fine and are being applied, and the copy is written for the settings editor — in Match Setup the user isn't changing settings, so "changes remain available for this session only" has no referent. Keeping the banner keyed on `status` and rendering storage warnings through their own (or a context-neutral) message would keep both surfaces honest.

**2. `save()`'s boolean return now conflates "not saved" with "saved but not cached".**

In the anonymous/local path, `save` became `return cacheAndCommit(...)`, so it returns `false` when only the *cache* write failed — even though the settings were parsed, committed, and are live in state. The other `false` returns mean validation failure, conflict, or a failed cloud write.

The current caller ignores the return, so nothing is wrong today. But `saveSoccerSettings` is exposed on `SettingsContext`, and a future caller treating `false` as "the user's change did not take" would be wrong in exactly this case. Either keep returning `true` when the change was applied and let `sync.error` carry the storage warning, or return a small result object so callers can distinguish.

---

### Test coverage

Good for the slice's scope, and the two storage cases (`contains browser storage read failures`, `reports browser storage write failures without throwing`) pin the specific behaviour that changed rather than the happy path.

The manual matrix in `REGRESSION_SOC_6D_SETTINGS.md` is the right artifact for what CI can't reach — RLS by role, two-session conflicts, and multi-account cache isolation. Worth confirming the matrix includes a cache-write-failure pass (quota-exceeded or private browsing), since that path is now user-visible and only unit-tested.

### Security

No authorization surface changes. Storage failures affect only the local mirror; every write still goes through the 048 CAS RPCs with their manager checks, and no fallback path lets local state substitute for cloud authority. That satisfies §8's "without weakening authorization" — I checked it as a claim rather than taking it from the description.

### Verdict

Approve. CI passes on the head commit.

That closes SOC-6D end to end. Both findings are small enough to fold in here or leave for 6E; neither affects correctness of the settings hierarchy or the cloud contract.

Rothermal:
Validated both findings and addressed them in 5377937.

1. Match Setup now distinguishes a successful team-default load with failed device caching from unavailable shared defaults. It displays context-specific copy: the defaults loaded and remain available for the session, but were not cached.
2. The local personal-settings save contract again returns true when parsed settings were applied to live state. A cache failure remains visible through sync.error, but no longer reports the applied change as an unsuccessful save.

Validation rerun: 97 test files / 613 tests passed, production build passed, and lint has 0 errors (the same 3 existing Fast Refresh warnings).

Rothermal:
## Re-review — `5377937`

Both findings fixed. Approve.

**Save contract.** `cacheAndCommit` returns `void` again and the local path calls it then `return true`, so a successful in-memory apply reports success while `sync.error` carries the storage warning. Dropping the boolean entirely is better than keeping it and ignoring it — now no caller can read a misleading result at all.

**Match Setup banner.** The storage case is special-cased to "Team defaults loaded, but they could not be cached on this device. They remain available for this session," which fixes the part I objected to: it states the defaults *did* load rather than implying they're unavailable. The sentinel comparison is sound, since both sides use the exported `SPORT_SETTINGS_STORAGE_ERROR` constant.

One passing observation, no action: matching on the message string is the kind of thing that rots if a second site ever produces the same user-facing text without the constant. A `code` on `SportSettingsCacheSaveResult` would be sturdier, but with one producer and one consumer it isn't worth churn now. Worth remembering if 6E adds more cache-backed surfaces.

### Verdict

Approve. CI passes on `5377937`.

That completes SOC-6D — schema and resolver, personal sync, team defaults with setup inheritance, and hardening. The manual matrix in `REGRESSION_SOC_6D_SETTINGS.md` is what stands between this and 6E; the RLS-by-role, two-session conflict, and multi-account isolation cases are the ones CI structurally cannot reach.

PR #247, opened 2026-07-27
Plan SOC-6E soccer release hardening

Rothermal:
## Summary
- document the approved SOC-6E availability and historical-access contract
- split delivery into 6E1 capability/availability, 6E2 regression hardening, and 6E3 production enablement
- define migration 049 as a narrow release-capability handshake
- record go/no-go, rollback, automated, and manual release boundaries
- correct stale SOC-0 and SOC-6 roadmap status

## Decisions
All eight focused Q&A decisions use the recommended option. Soccer remains development-only until 6E3 and remains disabled by default after release.

## Validation
Documentation-only change; git diff --check passed. Application tests were not rerun.


Rothermal:
Second-pass plan review completed in efe93ea.

Clarifications:
- 6E1 now replaces all scattered Soccer DEV availability guards with one centralized preview/released policy while production release stays off.
- 6E3 becomes a minimal production-policy flip and final sign-off rather than a risky route rewrite.
- Added an explicit surface matrix for unreleased, released/off, and released/on behavior.
- Existing and historical Soccer routes remain independent from release/toggle state; disabled dashboards retain history links but no New Game.
- Migration 049 is provisional pending the merged sequence and gates only new cloud creation, never history, recovery, settings repair, or local play.
- Capability errors distinguish missing backend, access failures, and retryable offline/network states; expected preflight failures do not create sync-error telemetry.
- Rollback returns to the 6E2 policy without reversing migration 049 or hiding existing data.
- Manual sign-off now requires recorded deployed commit, migration level, environment, browser/PWA mode, date, and reviewer.

Documentation-only update; git diff --check passed.

Rothermal:
## Plan review — SOC-6E soccer release hardening

Documentation-only. Reviewed the contract against the code it describes, since a plan's main failure mode is asserting something about the codebase that isn't true.

### What this does

Defines the availability and historical-access contract for releasing Soccer as an opt-in production sport, adds migration 049 as a capability handshake, and splits delivery into 6E1 policy/preflight, 6E2 regression evidence, 6E3 enablement.

**The §3 gap list is accurate**, which is worth stating explicitly because it's the part most likely to be stale:

- `isSportWorkspaceAvailable` really is `if (sportId === 'soccer') return development` — it ignores `enabledInSettings` for Soccer exactly as described (`sportAvailability.ts:1-7`).
- `App.tsx` gates Soccer setup, players, tracker, legacy review, and summary on `import.meta.env.DEV` (lines 44, 50, 56, 65, 72).
- `Admin.tsx:655,723` produce the "Development preview" / "Coming soon" labels.
- `Games.tsx:362` and `GameInfo.tsx:314` reject Soccer review in production.

`SportSelect.tsx:52`'s `isSoccerPreview` is a sixth gate the §3 narrative doesn't name, but §6's boundary list does include `SportSelect.tsx`, so nothing is missed.

Two design choices are right and worth keeping:

- **Separating release state from existing-record access is what makes rollback safe.** §6's rollback claim — redeploy the policy, no data migration, history stays reachable — only holds because §4.2 forbids historical routes from depending on release state or the device toggle. Those two clauses have to move together, and they do.
- **The capability cache is session-scoped and cleared on sign-out or account change**, carrying forward the cache-isolation lesson from SOC-6D2 rather than rediscovering it.

---

### Findings

**1. The centralized release policy has no stated answer for the other three configured sports.**

Today `isSportWorkspaceAvailable` is `if (sportId === 'soccer') return development; return enabledInSettings`, so baseball, football, and hockey are user-enablable through App Settings and appear in the sport list once toggled (`Admin.tsx:123,157` are the only call sites). §6 replaces that helper with "explicit released/discoverable/new-game and existing-record decisions" governed by "one centralized Soccer release-stage value" — but never says what stage those three sports carry.

If they inherit an "unreleased" default they silently disappear from Settings even when enabled, which is a regression for anyone using them today. If they're implicitly "released" then the policy is Soccer-specific in practice and should say so. §7's coverage line only promises *Basketball* availability is unchanged, so nothing in the plan would catch either outcome.

One sentence in §4.1 fixes it: non-Soccer sports keep `enabledInSettings` semantics, and add them to §7's unchanged-availability assertion.

**2. The capability error taxonomy has no client-stale direction, which the PWA makes the likelier one.**

§5.1 says the client "accepts only the exact supported contract version", missing RPC or schema-cache errors become `backend_update_required`, and "malformed or unsupported responses fail closed". That covers client-newer-than-backend. It doesn't cover backend-newer-than-client — which is what a cached PWA shell produces after a backend deploy, and §8 already lists "production PWA refresh" as a manual case.

In that state an older client sees a version it doesn't support, fails closed, and (by the taxonomy as written) reports *backend update required* — telling the operator to update a backend that is already ahead. That sends release-day debugging in the wrong direction.

Worth deciding explicitly in §5.1: either the client tolerates a newer contract version by ignoring unknown fields, or an unsupported-newer version maps to its own state with "reload the app" copy rather than reusing `backend_update_required`.

**3. Cloud-team Soccer setup will have two sequential preflights unless 6E1 sequences them.**

§5.2 requires capability preflight before a cloud-team flow may park or replace the active game. SOC-6D3 already added a blocking gate in the same flow: `handleContinue` refuses while `teamSettings.scopeTeamId !== selectedTeamId` or status is idle/loading, so Continue already waits on the team-defaults load.

Neither is wrong, but done independently the user gets two blocking waits in one Continue. Worth a line in 6E1's boundaries saying the capability check and the team-defaults load resolve together before Continue commits the cloud source, so the flow has one wait and one failure surface.

---

### Smaller notes

- §5.1 describes migration 049 as representing "the required 043-049 ... backend boundary", which includes itself. Reads oddly; "043 through 048 plus this handshake" is what's meant.
- `sportAvailability.test.ts` currently pins the soccer-uses-`development` behavior directly. 6E1 necessarily rewrites those assertions — worth naming in the slice so it isn't mistaken for an unrelated regression during review.
- `aggregateTransport.ts:355`'s `import.meta.env.DEV` guard is a development *log*, not an availability decision. §6's "Dev-only diagnostic tools remain environment-gated" and §7's "no Soccer feature component performs its own `import.meta.env.DEV` availability decision" are worded precisely enough to leave it alone — good, but a reviewer sweeping for DEV references will hit it, so it may be worth naming as an explicit exclusion.

### Verdict

Sound plan, and unusually well grounded — every §3 claim I checked matches the code, and the release/history separation is designed so that rollback is a redeploy rather than a data operation. Finding 1 is the one to settle before 6E1 starts, since it decides the shape of the policy API rather than an implementation detail. Findings 2 and 3 are cheap to resolve now and awkward to discover during release week.

Rothermal:
Validated all three findings and updated the plan in 9a0b5cd.

1. The release stage is now explicitly Soccer-only. Basketball, baseball, football, and hockey retain current enabledInSettings behavior, with regression cases required for all four.
2. Capability version direction is explicit: missing/older server -> backend_update_required; newer server/stale PWA client -> client_update_required with reload/update guidance; offline, access, and malformed states remain distinct.
3. Team defaults and release capabilities now load concurrently for the selected team behind one Continue wait and one failure surface.

Also incorporated the smaller notes: 049 now describes 043-048 plus the handshake, sportAvailability.test.ts is named as an intentional rewrite, and aggregateTransport.ts development logging is explicitly excluded from the availability-gate sweep.

Documentation-only update; git diff --check passed.

Rothermal:
## Re-review — `9a0b5cd`

All three findings and the three smaller notes are resolved. Approve.

**1. Release stage scoped to Soccer.** "Basketball, baseball, football, and hockey retain their current `enabledInSettings` availability behavior; SOC-6E does not assign those sports a release stage." Named explicitly, with a matching §7 coverage line and an unchanged-availability case for all four in the 6E1 test rewrite — so the outcome I was worried about (three sports quietly disappearing from Settings) is now something a test would catch rather than something a reviewer might notice.

**2. Version direction is explicit, and you picked the stricter option.** Five distinct states, each with the right operator direction: missing or older server → `backend_update_required`, newer server → `client_update_required` with reload guidance, auth/access unchanged, offline retryable, malformed fails closed. Adding "the client does not optimistically ignore fields from a newer contract" settles the forward-compatibility question rather than leaving it to the implementer — which is the part that would otherwise have been decided by whoever wrote the parser first.

**3. Concurrent preflight.** "Capability loading and the SOC-6D3 team-defaults load start together… Continue waits for both to reach a usable terminal state and exposes one combined loading/failure surface." That's exactly the shape, and §7 pins it.

The smaller notes all landed too: 049 now reads "043 through 048 … plus the 049 handshake", `sportAvailability.test.ts` is named as an intentional rewrite, and the aggregate-transport development log is explicitly excluded from the availability-gate sweep.

### One implementation note for 6E1, not a plan change

`client_update_required` guidance has to survive how this app actually updates. `vite.config.ts` uses `VitePWA({ registerType: 'autoUpdate' })`, so after a deploy the new service worker installs in the background and takes control on a subsequent navigation — meaning the *first* reload can still serve the old shell and reproduce `client_update_required`. Copy along the lines of "reload once more if this persists," or waiting on the SW update before prompting, keeps a user from reloading once, seeing the same error, and concluding the app is broken. The plan's "Update/Reload App guidance" is the right altitude; this is just the detail that makes it true in practice.

### Verdict

Approve. Ready for SOC-6E1.

The contract's strongest property is still the one from the first revision: because §4.2 keeps historical routes independent of release state and the device toggle, the §6 rollback really is a redeploy rather than a data operation. Worth protecting that invariant in review of 6E1, since it's the clause most likely to erode when the concrete route guards get written.

PR #249, opened 2026-07-28
SOC-6E1: add release policy and cloud capability preflight

Rothermal:
## Summary
- centralize Soccer preview, production release, discovery, new-game, and existing-record policy
- preserve existing local and cloud Soccer routes while production discovery remains disabled
- add migration 049 and strict account-scoped capability preflight for cloud-team starts
- check capabilities before parking or replacing games and provide an explicit local-only fallback
- update release documentation and add availability, migration, parser, cache, route, and mutation-order coverage

## Migration
Apply supabase/migrations/049_soccer_release_capabilities.sql before testing cloud-team Soccer starts. The production Soccer release flag remains off until SOC-6E3.

## Validation
- pnpm test: 100 files, 630 tests passed
- pnpm lint: passed with 3 existing Fast Refresh warnings
- pnpm build: passed


Rothermal:
## Code review — SOC-6E1 release policy and capability preflight

Reviewed against `PLAN_SOC_6E_RELEASE_HARDENING.md` §4-§7 (as revised in #247) and the existing availability/route code.

### What this does

Replaces the scattered Soccer `import.meta.env.DEV` availability checks with one `getSportAvailabilityPolicy`, adds migration 049 plus a typed capability parser/loader with session-scoped caching, and preflights cloud-team Soccer starts before any game mutation.

The three things I flagged during the plan review are all honored in code, and I checked each rather than taking the description:

- **Non-Soccer sports are genuinely unchanged.** `getSportAvailabilityPolicy` returns `discoverable`/`canStartNewGame` = `enabledInSettings` with `canAccessExisting: true` for anything that isn't Soccer, and `sportAvailability.test.ts` covers basketball, baseball, football, **and** hockey via `it.each`. That closes the gap where three sports could have silently vanished from Settings.
- **The rollback invariant holds in code, not just prose.** `canAccessExisting: true` is unconditional in both branches, and every Soccer route gate in `App.tsx` — setup, players, tracker, checkout redirect, summary, and `/soccer/review` — is *removed* rather than re-gated on the new policy. That's what makes §6's "rollback is a redeploy, not a data operation" true, and it was the clause I said was most likely to erode when concrete guards got written.
- **Prerequisites load concurrently behind one gate.** `teamDefaultsLoading` and `capabilityLoading` are computed independently and combined into `cloudPrerequisitesLoading`, with a single Continue disable and a single error surface — one wait, not two.

Other things worth noting because they're easy to get wrong:

- The version taxonomy matches the plan exactly: older → `backend_update_required`, newer → `client_update_required` with reload copy, plus distinct offline/access/invalid/not-configured states. `parseCapabilities` requires an exact key set *and* exact values, so the "does not optimistically ignore fields from a newer contract" clause is enforced structurally.
- The capability cache only stores `ready` results, is keyed by user id, and the in-flight dedupe clears itself only when `inFlight === request` — the same identity check that mattered in 6D2. `AuthContext` clears it on both account change and explicit sign-out.
- No-mutation-on-failure is real in `GameSetup`'s deep-link path: the `await ensureSoccerReleaseCapabilities` and its `if (cancelled) return` precede the `hasActiveGame` check, the confirm, and `startNewGame`. The static ordering tests back it up and fail closed — anchors are asserted present, and a missing `window.confirm` or `startNewGame` yields `-1`, which fails `toBeLessThan` rather than passing vacuously.
- Migration 049 is minimal and correctly locked down: definer with fixed `search_path`, `auth.uid()` and `has_active_app_access()` checks, revoked from `public`, granted only to `authenticated`, no product data, plus `notify pgrst, 'reload schema'` so the function is immediately visible.

---

### Findings

**1. Migration 049 does not verify the boundary it claims to represent.**

The RPC returns eight hardcoded literals. It proves that *049 itself* was applied — nothing else. But §5.1 says the contract "represents the required 043 through 048 Soccer transport, recovery, recorder, finalization, aggregate, and settings boundary", and the test is even named "returns the exact versioned 043 through 049 boundary".

That gap matters because migrations here are applied manually and selectively — the README tells operators to "run only the new ones". If 048 fails partway or is skipped and 049 is applied, preflight reports `ready`, the user proceeds into a cloud Soccer start, and the failure surfaces deeper as an opaque missing-function error from an operational RPC. That is precisely the outcome the preflight exists to prevent, and the client can't detect it because the server never reports anything but the constants.

Cheap fix that makes the handshake real: check one marker object per migration inside the function and fail closed if any is absent — e.g. `to_regclass('public.game_events')`, `to_regclass('public.game_event_canonical_publications')`, `to_regprocedure('public.get_soccer_scope_aggregate_publications(text,uuid,timestamptz,uuid,integer)')`, `to_regclass('public.user_sport_settings')`. Then extend `migration049.test.ts` to assert those probes exist, since the static suite is the only automated check on this SQL.

**2. One Retry permanently disables capability caching for the session.**

```ts
void ensureSoccerReleaseCapabilities(user.id, { force: capabilityAttempt > 0 })
```

`capabilityAttempt` only ever increments, so after the user clicks Retry once, *every* subsequent run of this effect passes `force: true` — and the effect re-runs on `teamSource`, `cloudAvailable`, and `user` changes. Toggling cloud → local → cloud then re-hits the RPC each time for the rest of the session, bypassing the cache the plan explicitly asks for.

Either track which attempt should force (compare `capabilityAttempt` against a ref of the last forced value) or drop `force` from the effect and have the Retry button call `ensureSoccerReleaseCapabilities(userId, { force: true })` directly.

**3. Error classification leans on message substrings that can mislabel real states.**

`isNetworkError` matches `'timeout'` / `'timed out'`, so a Postgres statement timeout — *"canceling statement due to statement timeout"* — would be reported as `offline`, telling a connected user they have no network. `isAccessError` matches `'jwt'`, so an expired token reports "Your account cannot start this cloud Soccer match" rather than a re-authentication state.

Both are unlikely for an RPC that returns constants, and supabase-js refreshes tokens on its own, so this is hardening rather than a live bug. Preferring `error.code` and reserving message matching for the cases where no code exists would keep the operator direction trustworthy — which is the whole point of having five distinct statuses.

---

### Smaller note

Following up my plan-review comment now that the copy is concrete: `client_update_required` reads *"This app is out of date. Reload or update the app before starting a cloud Soccer match."* With `VitePWA({ registerType: 'autoUpdate' })`, the replacement service worker installs in the background and takes control on a *subsequent* navigation, so the first reload can still serve the stale shell and reproduce the same error. Adding "if this keeps happening, reload once more" — or waiting on the SW update before surfacing the prompt — avoids a user concluding the app is broken. Worth handling in 6E2 alongside the PWA refresh matrix row.

### Test coverage

Good breadth for a policy slice: availability across every release/toggle combination with all five sports, capability parsing including older and newer contract versions, cache isolation by account, and source-ordering guards for all three preflight sites. The ordering tests are the right tool given these are cross-file invariants that unit tests can't express.

Gap matching finding 1: nothing asserts the handshake actually depends on 043-048 — because it doesn't yet.

### Security

Migration 049 is correctly scoped and adds no authority. The client never treats capability success as authorization: the preflight gates only new cloud-game creation, and operational RPCs keep their own RLS, CAS, and role checks. Historical reads, resume, recovery export, and local-only play remain outside the preflight, as §5.2 requires.

### Verdict

Strong slice — the availability policy is clean, the preflight ordering is correct at every entry point, and the historical-access separation that rollback depends on is properly implemented. Finding 1 is what I'd fix before merge: as written the handshake verifies its own presence rather than the contract it advertises, and hardening it is a few `to_regclass` calls. Findings 2 and 3 are small.

CI passes on the head commit.

Rothermal:
Reviewed each finding against the implementation and updated the PR in 0504d17.

- Finding 1 was valid. Migration 049 now probes the operational 043-048 tables and current RPC signatures. An incomplete boundary returns contract version 0, so the client reports backend_update_required without exposing missing object names. The migration contract tests now pin every probe.
- Finding 2 was valid. Retry now sets a one-shot force ref; the effect consumes and clears it, so later source/account renders return to the normal session cache.
- Finding 3 was valid hardening. JWT/PGRST301 failures now have an authentication_required state, and network text fallback is used only when no database/PostgREST code exists. Coded statement timeouts remain generic transport errors instead of being labeled offline.
- The PWA repeated-reload wording is deferred to SOC-6E2. That phase already owns service-worker refresh behavior and the deployed PWA refresh matrix, so the copy should be decided against the actual update lifecycle there.

Validation: pnpm test (100 files, 633 tests), pnpm lint (only the 3 existing Fast Refresh warnings), and pnpm build all pass.

Rothermal:
## Re-review — `0504d17`

All three findings fixed. Approve.

### Verified fixes

**1. The handshake now verifies the boundary it advertises.** The function probes nine tables and five RPC signatures before returning the real contract, and returns `contractVersion: 0` otherwise.

The risk this fix introduces is that a single wrong probe fails closed in *every* environment, so I checked all fourteen against the migrations rather than reading them for plausibility:

- `bind_soccer_event_game_v4(uuid,text,uuid,uuid,text,text,text,date,jsonb,jsonb)` matches the declaration at `046:700-711` and its own revoke at `046:1415-1417`;
- both aggregate RPC signatures match 047's revoke/grant exactly;
- both settings RPC signatures match 048's revoke/grant exactly;
- all nine tables resolve — `game_participants` and `game_event_stream_checkpoints` (043), `game_event_setup_snapshots` and `game_event_conflicts` (044), `game_event_primary_recorders` and `game_event_primary_recorder_audit` (045), `game_event_canonical_publications` (046), `user_sport_settings` and `team_sport_settings` (048).

Coverage reaches every migration in the claimed 043-048 range, so a skipped or half-applied step in any of them now degrades the contract instead of reporting `ready`.

Two details worth calling out. Returning `contractVersion: 0` rather than object names is the right instinct for a definer function — it gives the operator actionable direction without leaking schema. And it composes correctly with the existing client: `Number.isInteger(0)` passes the validity gate, the `< SOCCER_RELEASE_CONTRACT_VERSION` branch fires before the exact-shape check, so the single-key object becomes `backend_update_required` rather than `invalid_response`.

**2. One-shot force.** `forceCapabilityCheck` is consumed and cleared at the top of the effect before the async call, so exactly one forced load per Retry and later renders return to the session cache. Keeping `capabilityAttempt` as the re-run trigger while the ref carries the bypass intent separates "run again" from "skip the cache" cleanly, and the early consume means a re-render mid-flight can't re-arm it.

**3. Error classification.** `isNetworkError` now returns `false` whenever `error.code` is present — a better fix than pruning the two substrings I named, because it makes the entire class of coded-error-mislabeled-as-offline impossible. A statement timeout (57014) falls through to the generic transport error, and the new `authentication_required` state separates an expired session from an account that lacks access. The added test pins the coded-timeout case specifically.

### Deferral

Deferring the PWA repeated-reload wording to 6E2 is the right call — that phase owns the service-worker refresh behavior and the deployed PWA matrix, and the copy should be written against the observed update lifecycle rather than guessed at now.

### Verdict

Approve. CI passes on `0504d17`.

Worth carrying into 6E2: the probe list is now a hand-maintained mirror of migrations 043-048, and nothing fails if a future migration adds an object the handshake should require. The static test pins the probes that exist, not the ones that should. A line in the 6E2 matrix — or a note on the migration — reminding whoever adds migration 050 to extend the probe list would keep the handshake honest as the surface grows.

PR #250, opened 2026-07-28
SOC-6E2: Harden soccer release validation

Rothermal:
## Summary
- fail closed when legacy app settings contain malformed sport-toggle values
- contract-test the remaining direct development checks and improve stale installed-PWA recovery guidance
- add the consolidated SOC-6E release matrix and align release-status documentation

## Validation
- pnpm test (100 files, 640 tests)
- pnpm build

## Operator follow-up
The matrix intentionally leaves development/staging and unreleased-production browser evidence as Not run. Those passes are required before SOC-6E3 flips the production release flag.

No Supabase migration is added; migration 049 remains the release ceiling.


Rothermal:
## Code review — SOC-6E2 release hardening

Reviewed against `PLAN_SOC_6E_RELEASE_HARDENING.md` §6 (6E2 scope/exit) and the 6E1 policy it builds on.

### What this does

Fails closed on malformed legacy sport toggles, contract-tests the remaining direct development checks, improves the stale-PWA recovery copy, and adds the consolidated release matrix.

Small, well-scoped, and honest about what it does *not* finish. Things I verified rather than assumed:

- **The `mergeStoredSettings` rewrite is safe.** Dropping the `...parsed` spread would silently discard top-level keys, so I checked: `AppSettings` has exactly `enabledSports` and `courtCapture`, both are now constructed explicitly, and every consumer (`SettingsContext:35,43`, `ShotChartPanel:248`, `Admin:789+`) reads only those two. Filtering `enabledSports` to boolean values also preserves the previous behaviour of keeping unknown-but-boolean sport keys, which are inert because lookups go through `isSportEnabled`.
- **The development-check allowlist test is a stronger guard than it looks.** It walks all of `src/`, counts every `import.meta.env.DEV`, and asserts an exact per-file map — so a new direct check *anywhere*, including a file nobody thought to enumerate, fails the build. That enforces the plan's "no Soccer feature component performs its own availability decision" structurally rather than by convention, and the follow-up assertions naming `/dev/shot-chart` and the aggregate log document why the two survivors are allowed.
- **The matrix's Automated Evidence citations all resolve** — `gameParking.test.ts`, `settingsStorage.test.ts`, `sportAvailability.test.ts`, `releaseEntryGuards.test.ts`, `releaseCapabilities.test.ts`, `migration049.test.ts`, and the `src/hooks/` suites all exist on the branch. Worth checking, since this document is the sign-off artifact.
- **The matrix covers §8.** Sections A-G map onto the plan's list, and two cases stand out: E2-B03 exercises the 043-048 dependency probe from the last round (absent dependency in a disposable project → older contract, no schema details), and E2-B08 verifies that handshake success grants no operational authority. E2-B05 covers the stale installed-PWA case, and the new copy — "reload it… if this message returns in the installed app, close and reopen" — lands the guidance deferred from 6E1 against `registerType: 'autoUpdate'`'s activate-on-next-navigation behaviour.
- **Scope held.** No new Soccer product behaviour, no migration beyond 049. The settings change is app-wide input hardening, not Soccer logic.
- Also good hygiene: the stale "Soccer remains unavailable until SOC-6E" row in `REGRESSION_SOC_6D_SETTINGS.md` — which 6E1 invalidated when it removed the route gates — is corrected here rather than left to rot.

### Sequencing

Worth stating plainly since it governs what happens next: **this PR does not satisfy 6E2's exit condition.** The plan requires "manual results… recorded against a development/staging Supabase project through migration 049", and every operator row is `Not run`.

That's disclosed consistently rather than papered over — the PR body, the plan status line, a new implementation-status note above the unchanged exit condition, the README entry left as `- [ ]`, the codebase overview, and AGENTS.md all say evidence is pending. The original exit condition text was not weakened to match. That's the right way to land a partial slice, and it means 6E3 must not proceed on this alone.

### Notes

Neither is blocking.

**1. `statkeeper_settings` is now strictly schema-bound with no forward compatibility.** Unknown top-level keys are dropped, and `SettingsContext` persists the *merged* object on every change — so an older cached shell that loads settings written by a newer build will strip the newer field on the next write. Harmless today (two fields, both handled), and "fail closed" is the intent, but it's a real interaction with the stale-shell scenario this PR is partly about. Worth being deliberate before a third `AppSettings` field ever ships; a version marker or a passthrough for unrecognized keys would be the two options.

**2. The exact-count allowlist fails with a bare mismatch.** `toEqual({ 'src/App.tsx': 2, … })` is correctly strict for a release invariant, but a legitimate future dev-only addition — or removing the aggregate log — fails with a count diff and no explanation of the rule. A one-line comment ("extend only for dev-only diagnostics, never for availability decisions") would make that failure self-servicing.

### Verdict

Approve. CI passes on the head commit.

The remaining work is operator evidence, not code: the two preliminary passes (development/staging and unreleased production) against a project through migration 049. E2-B03 and E2-C07 are the two I would run first — both need a disposable project because they deliberately break the backend, and they're the cases most likely to surface something that changes code rather than just recording a result.

Rothermal:
Reviewed both non-blocking notes:

1. **Settings forward compatibility:** valid concern, deferred. The current strict schema boundary is intentional for SOC-6E2 fail-closed behavior, AppSettings currently has only the two explicitly reconstructed fields, and preserving unknown top-level input now would retain unvalidated data without a migration/version contract. Before adding a third top-level field, we should choose a versioned settings migration or an explicitly validated passthrough strategy.
2. **Allowlist failure guidance:** updated in commit 70a098a with an adjacent rule: extend the exact allowlist only for development diagnostics, never feature-availability decisions.

Affected guard test passes (6/6). No other review threads or blocking findings are present.

Rothermal:
## Re-review — `70a098a`

Both notes handled. Approve.

**Allowlist guidance.** The comment sits directly above the check and states the rule that matters — extend for dev diagnostics, never for feature-availability decisions. That converts a bare count mismatch into a self-servicing failure, which was the whole point.

**Forward compatibility, deferred with the decision recorded.** Agreed, and the reasoning is the right one: preserving unknown top-level keys today would retain unvalidated data with no version contract, which is a worse default than failing closed while `AppSettings` has exactly two explicitly reconstructed fields. Naming the trigger — choose a versioned migration or an explicitly validated passthrough *before* a third top-level field lands — is what keeps this from being rediscovered later. That was a non-blocking note and deferring it with a stated condition is the correct outcome, not a lesser one.

Also confirmed the middle commit is what it claims: a one-line matrix update flipping the implementation run's CI row from `Pending PR` to `Pass; PR #250`. No substantive change rode along.

### Verdict

Approve. CI passes on `70a098a`.

Unchanged from my last pass: this closes the code side of 6E2, not its exit condition. The two preliminary operator passes still have to be recorded in `docs/REGRESSION_SOC_6E_RELEASE.md` before 6E3 flips the release flag, and E2-B03 and E2-C07 remain the two worth running first — both need a disposable project and both can still surface something that changes code rather than just filling in a result.

PR #251, opened 2026-07-28
SOC-6E3: Enable soccer as a production opt-in

Rothermal:
## Summary
- enable Soccer in production through the single centralized release policy
- preserve the device-local default as off and pin both production toggle states in tests
- update SOC-0/SOC-6 status, migration guidance, operator evidence, and concrete rollback documentation
- require lint, tests, and the production build on the exact merged commit before Pages deploys

## Validation
- pnpm test (100 files, 641 tests)
- pnpm build (production PWA generated with 12 precache entries)

## Release decision
The owner approved an owner-only, post-deployment validation approach on 2026-07-28. The app currently has one user, no existing Soccer games, Soccer remains device-local opt-in and off by default, and rollback preserves existing records.

Merging deploys the release. The target project must be through migration 049. The detailed browser/PWA, role, cloud, multi-sport, and Basketball matrix remains the post-deployment iteration checklist and becomes a gate before access materially broadens.

The deploy workflow runs lint, tests, and build against the exact merged commit before publishing.

Rollback is code-only: restore SOCCER_RELEASED_IN_PRODUCTION and its shipped-value test assertion to false, run lint/test/build, and redeploy. Do not reverse migration 049 or alter existing Soccer records.

No Supabase migration is added.


Rothermal:
## Code review — SOC-6E3 production enablement

Reviewed against `PLAN_SOC_6E_RELEASE_HARDENING.md` §6 (6E3 scope and go/no-go) and the 6E1/6E2 policy.

### What this does

Flips `SOCCER_RELEASED_IN_PRODUCTION` to `true`, pins both production toggle states in tests, and updates release/rollback documentation.

The code change is exactly what the plan asked for — **one line**, in the centralized policy, with no route or component touched. That's the payoff for 6E1 doing the separation work properly, and it's worth noting that it actually held: nothing else needed to move.

Verified:

- **Off by default survives the flip.** The new test pins released-production with the toggle off to `discoverable: false, canStartNewGame: false, canAccessExisting: true`, and with it on to discoverable/startable. Existing-record access stays `true` in both, which is the invariant the whole rollback story depends on.
- **The matrix grew the right rows** — E3-A10/A11/A12 for released production with the toggle off/on/disabled-with-records, plus a deployed-Pages section covering base-path/HashRouter (E3-H01), OAuth return (E3-H02), stale installed-PWA refresh (E3-H03), offline cached shell (E3-H04), and a no-leak console/network check (E3-H05).
- **Documentation is consistent and does not overclaim.** Status lines, the 6E3 implementation note, the SOC-6 status, and the PR body all say operator evidence is pending, and the concrete rollback procedure is now written down.

---

### Findings

**1. Merging this PR publishes to production, and the go/no-go gate it is supposed to satisfy is entirely unrun.** *(the thing to settle before merge)*

`.github/workflows/deploy.yml` triggers on `push` to `stattracker` — this PR's base branch — so the merge itself builds and deploys to GitHub Pages. Two details make that sharper:

- the deploy job runs **only `pnpm build`** — no lint, no tests;
- `.github/workflows/ci.yml` triggers only on `pull_request`, so nothing re-runs the suite against the merged commit.

Meanwhile every operator row in `REGRESSION_SOC_6E_RELEASE.md` is `Not run` — not just the deployed pass, but the 6E2 preliminary development/staging and unreleased-production passes too. So the plan's own criterion —

> Go/no-go requires: … the complete manual release matrix signed off

— would be crossed with zero manual validation at any stage, and the only thing standing between merge and every user is a successful `vite build`.

I don't think this is intent; it's branch topology. The 6E3 note reframes sign-off as work recorded "before this phase is marked complete", which is true of the *phase* but not of the *deployment* — the deploy isn't gated on the phase marker, it's gated on the merge.

Three ways out, in ascending order of effort:

- record the 6E2 preliminary passes first (they're the ones the plan wanted before the flip), then merge;
- temporarily drop the `push` trigger from `deploy.yml` so publishing becomes deliberate — `workflow_dispatch` is already on that workflow, so the machinery exists;
- merge to a staging branch, validate the deployed artifact, then fast-forward.

This is your release call, not mine. I'm flagging the mechanism because the documentation reads as "pending sign-off" while the merge button is the release.

**2. The documented rollback would fail its own PR's CI.**

The plan and PR body both say: set `SOCCER_RELEASED_IN_PRODUCTION` back to `false`, build, redeploy. But `sportAvailability.test.ts` now asserts `expect(SOCCER_RELEASED_IN_PRODUCTION).toBe(true)`, and the two behavioural assertions in that test rely on the module default rather than passing `soccerReleasedInProduction` explicitly — so a rollback flips three assertions red.

The deploy itself would still succeed (it doesn't run tests), but the rollback *PR* would be red, and CI-red-during-an-incident is exactly when you don't want an unexplained failure. Two clean options: document the rollback as `git revert` of this commit, which reverts both files together; or pass `soccerReleasedInProduction` explicitly in the two behavioural assertions so only the single value assertion needs touching. Pinning the shipped value is worth keeping either way — it forces the flip to be deliberate.

**3. Note: the merged commit is never tested.**

Following from finding 1 — `ci.yml` is `pull_request`-only and `deploy.yml` builds without testing, so no workflow runs `pnpm test` against what actually ships. Fine while merges are fast-forwards of a green PR, but a semantic conflict between two PRs merged in sequence would deploy untested. Adding `push: branches: [stattracker]` to `ci.yml`, or a test step to the deploy job, closes it. Out of scope for this PR; worth a follow-up given a release is about to ride these workflows.

---

### Verdict

The code is correct and minimal, and the one-line diff is evidence that 6E1's separation was done right.

My reservation is sequencing, not implementation: merging is deploying, and the manual matrix the plan makes a go/no-go condition has no rows recorded at any stage. If you're satisfied that the automated coverage plus a post-merge operator pass is an acceptable risk for an opt-in, off-by-default sport, that's a legitimate call — the blast radius is genuinely small, since discovery stays off until each device opts in and existing records were separated from release state back in 6E1. I'd still record the two preliminary passes first, because they're cheap relative to validating a live build.

CI passes on the head commit.

Rothermal:
Validated all three findings and updated the PR in commit

Rothermal:
## Re-review — `0e72d9d`

All three findings addressed, and finding 1 is solved better than what I proposed.

### Verified fixes

**1. Release gate.** You kept the `push` trigger and gated at the PR instead — and I confirmed the PR is actually `isDraft: true`, not merely documented as such. That matters: GitHub refuses to merge a draft, so the gate is enforced by the platform rather than by remembering to revert a temporarily-edited workflow. My suggestion would have left a modified `deploy.yml` that someone had to remember to restore; this doesn't. The matrix now carries the rule explicitly — "keep its PR in draft and treat the build as an unsigned release candidate" — so un-drafting becomes the deliberate release action the go/no-go list describes.

**2. Rollback no longer fights its own tests.** The behavioural assertions now pass `soccerReleasedInProduction: true` explicitly, and the shipped-value check is isolated to a single line. A rollback touches exactly two lines, and the plan's procedure was updated to name the assertion and the lint/test/build run. That's the cleaner of the two options I offered.

**3. The deployed commit is now tested.** `deploy.yml` runs `pnpm lint` and `pnpm test` before `pnpm build`, so the exact merged commit is validated rather than just built. I checked that this won't break on secrets: `ci.yml` already runs the same three steps with no env vars, and only the build step needs `VITE_SUPABASE_*`.

### One interaction worth knowing

Fixes 2 and 3 combine to raise the stakes slightly on the rollback path: now that deploy runs the suite, a rollback that flips the flag without updating the shipped-value assertion will fail at deploy time, not just in its PR. The plan documents exactly that sequence, and the isolated one-line assertion makes it trivial — but under incident pressure `git revert` of the release commit remains the fastest route, since it takes both files atomically and can't leave the pair inconsistent. Might be worth naming as the preferred procedure with the manual edit as the fallback.

### Verdict

Approve the code. CI passes on `0e72d9d`.

The release decision is now properly sequenced: the draft status blocks the deploy until someone records the operator evidence and deliberately marks this ready. That was my whole concern, and it's resolved by a mechanism rather than by intent.

What's left is operator work, not review work — the preliminary development/staging and unreleased-production passes, then the deployed Pages pass with E3-H01 through E3-H05. Of those, E3-H02 (OAuth return to `/cursor-default/`) and E3-H03 (installed PWA across a deploy) are the two that can only be proven against the real deployment, so they're the ones the draft gate genuinely exists for.

Rothermal:
Release decision updated: the owner approved merging as an owner-only, default-off opt-in rollout with post-deployment validation. The repository now records that acceptance and keeps the full matrix as the iteration checklist and as a gate before access materially broadens.

The safeguards from the prior review remain in place: migration 049 capability preflight, explicit rollback, and lint/test/build on the exact merged commit before Pages deployment. I will mark the PR ready once refreshed CI passes on this documentation update.

PR #252, opened 2026-07-29
Revise BKE-0 after completed Soccer rollout

Rothermal:
## Summary
- update BKE-0 against the completed SOC-6 authority, summary, aggregate, settings, and release architecture
- correct Basketball-specific event contracts for assists, score adjustments, staff actors, and participant resolution
- split BKE-1 into 1A-1C and BKE-4 into 4A-4E, with settings/rollout in BKE-5 and clock/lineups in BKE-6
- add the required 24-question Basketball product-model Q&A for the future full BKE planning branch
- synchronize the roadmap, README, and agent codebase overview

## Validation
- git diff --check
- local Markdown link targets verified

## Follow-up
The product-model Q&A is intentionally not answered in this PR. It will be completed after merge on a full BKE planning branch before BKE-1A begins.


Rothermal:
## Plan review — BKE-0 revision after the completed Soccer rollout

Documentation-only. I reviewed the original BKE-0 across three rounds and every SOC-6 slice, so I checked two things specifically: whether the SOC-6 claims match what actually shipped, and whether the decisions settled in the first review were preserved or deliberately changed.

### What this does

Rewrites BKE-0 against the completed SOC-6 program, reverses the assist model, splits BKE-1 into 1A-1C and BKE-4 into 4A-4E, moves settings/rollout to BKE-5 and clock to BKE-6, and adds a required 24-question Basketball product-model Q&A that this PR deliberately leaves unanswered.

**The assist reversal is the substantive change, and it is right.** The original decision — assist as an actor on the made shot — was borrowed from Soccer, and §4.3 now names why that doesn't transfer: Basketball's stat grid records assists independently, so an actor-only model makes a standalone `ast` entry impossible or forces a fabricated shot. That's a genuine domain difference, not a preference.

It is also propagated completely. I checked all seven places the old model appeared:

- catalog gains `basketball.assist` with `relatedEventId` and an `assister` actor, and the shot row drops the actor — with an explicit "there is **no** `assist` or `blocked_by` actor on the shot";
- coverage table maps `ast` → `basketball.assist`;
- validation adds the link rules (active made shot, assister ≠ shooter, same `teamSide`);
- §6.1's `ast` row becomes "tombstone the newest matching assist; a linked shot remains untouched", and the shot row now tombstones linked assists too;
- §6.3 clear-chart adds linked assists to the removed set, correctly matching today's `['ast','oreb','dreb']` reversal while still excluding blocks;
- §8's F13 row and §11's decision row both updated.

Other things I verified rather than took on trust:

- **§3.6's six SOC-6 lessons are accurate.** The discriminated `local | cloud_primary | cloud_recording | canonical` authority, read-only remote/canonical sources, diagnostic suppression of official totals, TypeScript rebuild of canonical publications rather than treating `game_stats` as a second authority, layered settings with an immutable snapshot, and release/history separation all match what shipped in 6A-6E.
- **The phase renumber is clean.** No stale references — the opt-in consistently says BKE-5 across §2, §3.3, §10, and §11 row 11, and the clock consistently says BKE-6 including "until BKE-6, `min` stays a manual stat".
- **Stated product rules are actually enforced in the contract.** §11 row 4 says official corrections require a note, and §4.5's validation row encodes `official_correction` requires a non-empty `note`. That's the failure mode I've flagged on other plans — a rule asserted in prose but absent from the contract — and it isn't present here.
- The new "Catalog status" note marking the catalog as not frozen until the Q&A is a good correction; the earlier version presented it as more settled than it was.

The BKE-1 split is also an improvement on what I approved: 1A proves Soccer is unchanged, 1B proves a deterministic Basketball foundation with no UI, 1C proves visible court parity — one proof per PR, with the atomic mutation that F13 and clear-chart depend on landing first in 1A.

---

### Findings

**1. The reversal drops a benefit the original decision was chosen for, and the Q&A doesn't ask about it.**

The old §4.3 justified actor-on-shot partly like this: F7's "Made → Assisted by?" is *one user action* that today becomes two undo rows, and one event with two actors "makes F12's double row disappear."

With two events, that problem returns — and §6 still specifies single-event undo: "Undo tombstones the newest active event", with F12 keeping "strict newest-first removal through BKE-2". So in BKE-1C and BKE-2, undoing one assisted-shot capture takes two taps, and a single tap leaves either an orphaned assist or a shot whose assist still counts. §4.3 answers this only for BKE-3 ("the Timeline may present that linked pair as one play"), which is two phases later.

The two goals aren't actually in tension. §6.2 already establishes that a grouped removal undoes as one atomic mutation and that dependents are recoverable through their retained `relatedEventId`; the symmetric rule — an atomic multi-event *append* reverses as one unit — is a small addition to the command layer, not a model change.

Worth resolving in the plan rather than discovering it in BKE-1C acceptance: either state that F12 treats an atomic command group as one undo unit (and that the command layer records the group), or state plainly that an assisted shot costs two undo taps until BKE-3 and add that to §13's regression expectations. Either way, §12b q19 currently only asks to *confirm* newest-first — it should also ask what "newest" means for a pair appended by one command.

**2. Dangling reference in §6.1.**

The closing paragraph still reads "any decrement whose contract row says **'revise'** or names a multi-event consequence…". After the reversal no row says "revise" — the `ast` row tombstones now. Drop the clause so the sentence describes the table it follows.

---

### Sequencing

Worth stating plainly, as with the recent release PRs: **merging this does not approve BKE-0.** §9's BKE-0 exit condition requires the product catalog and contracts approved, and §12b's 24 questions are deliberately unanswered. The PR body says the same, so this is staging the questions rather than closing the phase — BKE-1A stays gated until the Q&A is completed on the planning branch.

### Validation

Documentation-only, so there's no code, test, or security surface to assess. `git diff --check` plus link verification is the right bar, and CI passes. The one thing worth confirming before merge is that the four `docs/PLAN_BKE_*` filenames introduced in §9 don't collide with anything already committed, since they're now referenced from README and the codebase overview.

### Verdict

Strong revision. The SOC-6 integration is faithful, the phase splits encode experience rather than repeating discovery, and the assist reversal is well-reasoned and thoroughly propagated. Finding 1 is worth settling before the Q&A pass, because it's a product-behavior question that shapes BKE-1C's acceptance criteria; finding 2 is a one-line cleanup.

Rothermal:
Reviewed both findings and updated the PR in commit 1e9a2ae.

- **Grouped undo: valid.** Multi-event Basketball capture payloads now carry an immutable captureCommandId. F12 treats the newest capture command as one durable undo/restore unit, while 
elatedEventId continues to own semantic linkage. This survives reload without widening the generic event envelope or inferring groups from adjacency/timestamps. Q&A question 19 and the regression mapping now state the one-tap behavior explicitly.
- **Dangling 
evise wording: valid.** Removed the obsolete clause from the decrement contract.
- **Filename check:** no collisions. PLAN_BKE_0_BASKETBALL_EVENT_ARCHITECTURE.md is currently the only committed docs/PLAN_BKE_*.md; all proposed phase filenames remain available.

git diff --check passes. The 24-question product-model Q&A remains intentionally deferred, so merging this PR still does not approve BKE-0 or unblock BKE-1A.

Rothermal:
## Re-review — `1e9a2ae`

Both findings resolved, and the grouped-undo answer is better than either option I offered.

### Verified fixes

**1. `captureCommandId`.** I suggested either defining grouped undo or documenting the two-tap cost. You did the former, and the design holds up on the details that usually break this kind of thing:

- **It's Basketball-owned payload metadata**, not a generic-envelope or column change — so the shared platform stays sport-neutral and BKE-1A doesn't have to widen the envelope for a Basketball product decision.
- **It's durable rather than inferred.** The explicit "never infers grouping from adjacency, timestamps, or `relatedEventId`" is the important sentence: adjacency-based grouping silently breaks after reload or once an intervening event lands, which is exactly the case F12 has to survive.
- **The two concerns stay separate.** `relatedEventId` owns semantic linkage, `captureCommandId` says only "one user gesture" — so a later manually-linked assist doesn't get absorbed into an earlier capture. The relationship rule spells that out ("later link/edit commands preserve each event's original group and never absorb events into it"), and the "previously unused" clause blocks the other direction, grafting onto an existing group by reusing an id.
- **Immutability is validated** — null or a non-empty generated id, unchangeable after revision 1 — so an edit can't move an event between groups after the fact.
- Undo resolves by id rather than position, so a later `elapsedMs` edit that reorders members doesn't fracture the group, and restore is symmetric with §6.2's grouped-restore rule.

Net effect: the double-undo-row problem that originally motivated actor-on-shot is solved by grouping instead, without giving up standalone stat-grid assist capture. That's strictly better than either prior model rather than a trade between them.

q19 now states the behaviour instead of asking to confirm a bare "newest-first", and both §13 and the roadmap's regression list name one-tap undo/restore for capture-command groups.

**2. Dangling "revise" clause** removed from §6.1.

**Filename check** independently confirmed: `PLAN_BKE_0_BASKETBALL_EVENT_ARCHITECTURE.md` is the only committed `docs/PLAN_BKE_*` file on `stattracker`, so all fourteen proposed phase filenames are free.

### Verdict

Approve. CI passes on `1e9a2ae`.

Sequencing unchanged and correctly stated in your update: merging this stages the questions rather than closing the phase — §12b's 24 answers and BKE-0 approval still gate BKE-1A.

One thing worth carrying into that Q&A pass: `captureCommandId` is now specified as available on *every* Basketball payload, so BKE-1B will want a shared payload base rather than repeating the field across thirteen-plus event definitions and their validators. Implementation detail, not a plan change — the cross-cutting validation row already states it correctly.

